PR #2, opened 2016-08-23
[WIP] Function test for entire client based upon info from an HMC extraction tool

andy-maier:
Updated on 2016-10-07.

Ready for review and merge.

This MR provides an HMC extraction tool `tests/function/extract_hmcinfo`, that extracts information from an HMC and writes it to an HMC info file in JSON format.

It also provides a new function test case `tests/function/run_hmcinfo.py` which runs operations against an HMC and that compares the actual results with the expected results in an HMC info file (e.g. as generated by the HMC extraction tool).

**How to try it out:**
- Extract data from the HMC by running the extract tool, into `tests/function/run_hmcinfo.json` (the file must have the same base name and directory as the function test case):
  
  ```
  PYTHONPATH=. tests/function/extract_hmcinfo <hmc> -u <hmcuser> -o tests/function/run_hmcinfo.json
  ```
- The function test case gets its input parameters from a config file. Copy the example config file `tests/function/example_config.yaml` to`tests/function/config.yaml`, and update it for your HMC.
- Run the function test case:
  
  ```
  py.test tests/function/run_hmcinfo.py -s
  ```

### DIFF
--- a/tests/function/_hmcinfo.py
+++ b/tests/function/_hmcinfo.py
@@ -1,0 +1,132 @@
+# Copyright 2016 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Encapsulation of the HMC info file.
+
+This module is used for testing only. It is contained in the main package
+subtree in order for it to be found for importing.
+
+An *HMC info file* contains the input and output to HMC operations at a level
+that can be used to verify these operations, and to set up the mock environment
+for the operations.
+
+An HMC info file can be produced with the `tools/extract.py` script and is
+consumed by the `tests/test_hmc.py` test case.
+"""
+
+from __future__ import absolute_import
+
+import json
+from datetime import datetime
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
+import pytz
+
+from zhmcclient import Error
+
+
+class HMCInfo(object):
+    """
+    The content of an HMC info file.
+    """
+
+    def __init__(self, hmc, userid):
+        """
+        Parameters:
+          hmc (string): Host name or IP address of HMC
+          userid (string): Userid on HMC that was used to extract the data
+        """
+        dt = datetime.now(pytz.utc).isoformat()  # timezone-aware UTC time
+        ops = OrderedDict()  # key: method + uri, value: see add_op()
+        self._data = OrderedDict()
+        # The following kind of initialization preserves the order:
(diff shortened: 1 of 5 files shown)


GitCop:

 * Commit: acc5f60d98eb007a6d976da32007cab4878c9da8
  * Invalid signoff. Commit message must end with  
    **Signed-off-by: .* <.*>**

---

andy-maier:
+1

GitCop:

 * Commit: 3291b53ba010b18dcc8a2c75ad371876dcb14cfe
  * Your subject line is longer than 80 characters

---

andy-maier:
Rebased this PR, in order to benefit from the latest changes on master.

leopoldjuergen:
For my understanding the hmc_extract.py script is a script to get the inventory of all resources managed by the zHMC. I would propose to rename the script to get_inventory.py. 
I don't like that we put any test code into the zhmcclient source folder like 
the _tests folder. We should separate source and test code.

andy-maier:
* On putting the `hmcinfo` module into the `_tests` subdirectory of the `zhmcclient` package: It is imported by the `hmc_extract.py` script, so it needs to be somewhere in the module search path. We can talk about renaming the `_tests` directory to something else; it was named that way because the `hmc_extract.py` script is a tool that is used in context of function testing.

* On the name of the `hmc_extract.py` script: Conceptually, it is not limited to GET operations or to building an inventory. It is meant to find out how the HMC behaves for all kinds of operations, including POST/DELETE, and operations with incorrect input that will fail. Currently, only GET operations on valid URIs are implemented. Given this, what would be a good name for the script?

andy-maier:
Rebased and squashed all commits into one.

andy-maier:
+1

andy-maier:
Addressed Juergen's comments from aug 26 as follows:

1. Name of `hmc_extract.py` script:

   > For my understanding the hmc_extract.py script is a script to get the inventory of all resources managed by the zHMC. I would propose to rename the script to get_inventory.py.

   Its main purpose is not to get an inventory, but to create an HMC info file. So I renamed it to `create_hmcinfo.py`.

2. `zhmcclient/_tests` directory:

   > I don't like that we put any test code into the zhmcclient source folder like the `_tests` folder. We should separate source and test code.

   I moved its content (the `_hmcinfo.py` file) into the `tests/function` directory which is a package anyway because py.test requires that. The only drawback is now that it needs to be imported from `tests.function`, which can be confused with a Python package named `test`. Note that relative imports are allowed only from within a package, and the scripts that need this module are not modules in packages.

andy-maier:
What is still needed for this PR, is to utilize the new mock support of the zhmcclient. Unfortunately, this will change quite a bit in this PR.

One approach would be:
- a script that creates the resource dictionary that is needed for mocking, as a JSON file, from an HMC, using the low level `Session.get()` operation. This file can lateron be used for setting up the mocking and it also defines expected results for retrieval operations. This script is similar to `create_hmcinfo.py`, just the output format is different.
- a function test that reads the resource dictionary JSON file of a particular HMC and that drives operations against that HMC, and compares the expected results. This script is similar to `run_hmcinfo.py`, except that it cannot generically rely on the operations in the file, but needs to have its own understanding of the operations. 
- a unit test that sets up zhmcclient mocking based upon a resource dictionary JSON file and that drives the function test mentioned above against the mocked environment.

andy-maier:
With the new focus on end2end testing and its fixtures, the extraction approach fro this PR is no longer needed. CLosing the PR.

PR #17, opened 2016-08-24
Fixed all flake8 issues and enforced issue-free flake8 in make check.

andy-maier:
This PR removes the ignoring of the flake8 error code in `make check`, so that an issue-free flake8 is now enforced. Any future flake8 issues will cause the Travis CI build to fail.

Because PyLint messages are harder to map to good or bad, PyLint issues are still ignored in `make check`, and PyLint issues need to be resolved on a voluntary basis.

Ready for review and merge.


leopoldjuergen:
+1

PR #21, opened 2016-08-24
Addressed PyLint issues and added uri property to BaseResource class.

andy-maier:
The `uri` property was added for two reasons:
- in order to avoid access to the private `_uri` instance variable in the unit tests.
- because it is a rather important attribute for the resource and was not available otherwise yet.

ready for review and merge


andy-maier:
+1

leopoldjuergen:
+1

PR #31, opened 2016-08-25
Sorted the operations in TimeStats display.

andy-maier:
Ready for review and merge.


andy-maier:
It turns out the reformatting of the lines with `.format()` does not help the coverage tool to recognize the lines as executed. Maybe a bug in the coverage tool. Maybe it has to do with the backslash as a line continuation.

Anyway, from a functional perspective, my vote is +1.

leopoldjuergen:
+1

PR #35, opened 2016-08-29
Fixed incorrect args/kwargs prototype generated when using _log_call

andy-maier:
This fixes the documentation (although code is changed).

Ready for review and merge.

Details:
- The documentation generated for methods or functions decorated with
  the `_log_call` decorator is incorrectly generated with a prototype
  of `*args, **kwargs`. This the prototype of the wrapper function
  of the decorator. This has been fixed by switching from using
  the `functools` package to using the `decorator` package.
  See also http://stackoverflow.com/a/14448445/1424462
  and https://pythonhosted.org/decorator/documentation.html#a-trace-decorator

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,5 @@
 six # MIT
 pbr>=1.8 # Apache-2.0
 requests>=2.10.0 # Apache-2.0
+decorator # new BSD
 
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #38, opened 2016-08-30
Add string representation of BaseResource.

leopoldjuergen:
Details:
- Added __str__ and __repr__ to BaseResource class.
- Added unit test.

### DIFF
--- a/zhmcclient/_resource.py
+++ b/zhmcclient/_resource.py
@@ -174,3 +174,25 @@ class BaseResource(object):
                 raise
             self.pull_full_properties()
             return self._properties[name]
+
+    def __str__(self):
+        """
+        Convert a BaseResource object to a string.
+
+        example:
+        Cpc(name=P0000S12,
+        object-uri=/api/cpcs/f1bc49af-f71a-3467-8def-3c186b5d9352,
+        status=service-required)
+        """
+        properties_keys = self._properties.keys()
+        search_keys = ['status', 'object-uri', 'element-uri', 'name']
+        sorted_keys = sorted([k for k in properties_keys if k in search_keys])
+        info = ", ".join("%s=%s" % (k, self._properties[k])
+                         for k in sorted_keys)
+        return "%s(%s)" % (self.__class__.__name__, info)
+
+    def __repr__(self):
+        """
+        Convert a BaseResource object to a string.
+        """
+        return self.__str__()
--- a/tests/test_resource.py
+++ b/tests/test_resource.py
@@ -87,6 +87,7 @@ class InitTests(ResourceTestCase):
         self.assert_properties(res, init_props)
         self.assertTrue(int(time.time()) - res.properties_timestamp <= 1)
         self.assertEqual(res.full_properties, False)
+        self.assertTrue(repr(res).startswith(res.__class__.__name__ + '('))
 
     def test_simple(self):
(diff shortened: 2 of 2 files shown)


leopoldjuergen:
+1

zhmcclient/_resource.py
```diff
@@ -174,3 +174,29 @@ def get_property(self, name):
...
+
+        example:
+        <Cpc _full_properties=False, _image_activation_profiles=None,
+        _load_activation_profiles=None, _lpars=None,
+        _manager=<zhmcclient._cpc.CpcManager object at 0x7f439cab3510>,
+        _partitions=None, _properties={u'status': u'service-required',
+        u'object-uri': u'/api/cpcs/f1bc49af-f71a-3467-8def-3c186b5d9352',
+        u'name': u'P0000P01'}, _properties_timestamp=1472558081,
+        _reset_activation_profiles=None,
+        _uri=/api/cpcs/f1bc49af-f71a-3467-8def-3c186b5d9352>
```
andy-maier: This is not the usual syntax used for `__str__()` or  `__repr__()`.

The usual syntax used for `__repr__()` is a Python constructor syntax, e.g.:

    Cpc(_full_properties=False, _image_activation_profiles=None, ...)

Because `__str__()` is for human consumption, I don't think it makes a lot of sense to show all 100+ properties there. Maybe just name, status and object-uri?

zhmcclient/_resource.py
```diff
@@ -174,3 +174,29 @@ def get_property(self, name):
                 raise
             self.pull_full_properties()
             return self._properties[name]
+
+    def __str__(self):
+        """
+        Convert a BaseResource object to a string.
+        Used for printing.
```
andy-maier: "Used for printing" does not really define the purpose. "Used for human consumption" would be better, because that makes it clear that users are not supposed to parse the result.

zhmcclient/_resource.py
```diff
@@ -174,3 +174,29 @@ def get_property(self, name):
...
+        _reset_activation_profiles=None,
+        _uri=/api/cpcs/f1bc49af-f71a-3467-8def-3c186b5d9352>
+        """
+        sortedkeys = sorted(self.__dict__.keys())
+        info = ", ".join("%s=%s" % (k, getattr(self, k)) for k in sortedkeys)
+        return "<%s %s>" % (self.__class__.__name__, info)
+
+    def __repr__(self):
+        """
+        Convert a BaseResource object to astring.
```
andy-maier: typo: "astring".

zhmcclient/_resource.py
```diff
@@ -174,3 +174,29 @@ def get_property(self, name):
...
+        Used for printing a list containing a BaseResource object.
```
andy-maier: "Used for printing" does not really define the purpose. "Used as debug information" would be better, I think.

Also, this is a single resource. Not sure I understand why it says "a list containing a BaseResource object"?

andy-maier:
My only remaining comment is on `__repr__()`:

* What do we want to be the purpose of `__repr__()`? My suggestion is that should be helpful for debugging, which implies that it should show the instance variables of the object. That is not the case in the code as of commit f6f6a69, where it simply invokes `__str__()`. By the way, I like `__str__()` as it is now.

PR #48, opened 2016-09-01
Add NIC support

leopoldjuergen:
Details:
- Add NicManager class.
- Add Nic class.
- Add Unit Tests for NicManager and Nic.
- Add NicManager to Nic class.
- Add NIC to documentation reference.
- Add NIC example (example7.py).

### DIFF
--- a/zhmcclient/_nic.py
+++ b/zhmcclient/_nic.py
@@ -1,0 +1,183 @@
+# Copyright 2016 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A **NIC** object represents a single Network interface card (NIC)
+of a physical z Systems or LinuxONE computer that is in DPM mode
+(Dynamic Partition Manager mode).
+Objects of this class are not provided when the CPC is not in DPM mode.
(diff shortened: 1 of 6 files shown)


zhmcclient/_nic.py
```diff
@@ -0,0 +1,183 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A **NIC** object represents a single Network interface card (NIC)
+of a physical z Systems or LinuxONE computer that is in DPM mode
+(Dynamic Partition Manager mode).
```
andy-maier: This sentencer suggests that the term NIC is used for the physical network adapter. The last sentence of this paragraph describes that better, and could be used as the definition of what a NIC is.

zhmcclient/_nic.py
```diff
@@ -0,0 +1,183 @@
...
+        Parameters:
+
+          properties (dict): Properties for the new NIC.
+            See the section in the :term:`HMC API` about the specific HMC
+            operation and about the 'Create NIC' description of the members
+            of the passed properties dict.
+
+        Returns:
+
+          string: The resource URI of the new NIC.
```
andy-maier: Depending on the outcome of the discussion what to return here that we started for the Adapter resource, we should change the return value here as well.

zhmcclient/_nic.py
```diff
@@ -0,0 +1,183 @@
...
+    def create(self, properties):
+        """
+        Create and configures a NIC with the specified resource properties.
+
+        Parameters:
+
+          properties (dict): Properties for the new NIC.
+            See the section in the :term:`HMC API` about the specific HMC
+            operation and about the 'Create NIC' description of the members
+            of the passed properties dict.
```
andy-maier: See my comment about phrasing such a reference, in the Adapters PR #45.

andy-maier:
+1 - we agreed to defer the comments until later.

leopoldjuergen:
+1

PR #58, opened 2016-09-08
Update example5

leopoldjuergen:
Details:
- Adapt example5 after the return resource object
  instead of URI changes.


andy-maier:
+1

PR #64, opened 2016-09-08
Updated documentation of resources classes, part 2.

andy-maier:
Please review. Only docs changes.

Details:
- Moved description of resource and manager base classes into appendix.
- Made terms in Resource Model glossary consistent and fixed any links between them.
- In all resource and manager descriptions, turned all references to resources into links to their entries in the resource model glossary.
- Removed `__init__()` methods of resource classes and manager classes from the documentation, because these objects are not supposed to be created by the user.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -23,6 +23,26 @@ This section contains information that is referenced from other sections,
 and that does not really need to be read in sequence.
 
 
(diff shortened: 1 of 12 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #65, opened 2016-09-09
Fixes #62: Changed get_connected_vnics() to return Nic objects

### LINKED ISSUE #62: get_connected_vnics() returns NICs as URIs
`VirtualSwitch.get_connected_vnics()` returns the NICs that are connected to the virtual switch.
The current design returns these NICs with their URIs. This is inconsistent to other list functions that return resources as their corresponding Python object (`Nic` in this case).

Proposal: Return the connected NICs as a list of `Nic` objects that have just their `object-uri` property set.

andy-maier:
Please review.

The code to create the `Partition` and `Nic` objects has now been consolidated into the respective resource modules, so that the `VirtualSwitch.get_connected_vnics()` method that needs them is now cleaner.

### DIFF
--- a/tests/test_nic.py
+++ b/tests/test_nic.py
@@ -230,5 +230,24 @@ class NicTests(unittest.TestCase):
             status = nic.update_properties(properties={})
             self.assertEqual(status, None)
 
+    def test_nic_object(self):
+        """
+        This tests the `nic_object()` method.
+        """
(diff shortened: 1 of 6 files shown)


andy-maier:
Rebased.

andy-maier:
+1

andy-maier:
Rebased.

PR #73, opened 2016-09-14
Removed downloads button in readme and added bigquery in comment.

andy-maier:
PyPI download statistics are broken at the moment, see for example [this issue](https://github.com/badges/shields/issues/716). Google BigQuery can be used instead, see [this article](https://mail.python.org/pipermail/distutils-sig/2016-May/028986.html) and these [example bigqueries](https://gist.github.com/alex/4f100a9592b05e9b4d63).

This PR removes the PyPI downloads/months button and adds a BigQuery for querying the package downloads by version in the comment(!) of the README.rst file.

Please review and merge.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,13 +16,39 @@
 zhmcclient - A pure Python client library for the z Systems HMC Web Services API
 ================================================================================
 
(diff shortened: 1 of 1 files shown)


leopoldjuergen:
+1

PR #74, opened 2016-09-14
Renamed get_connected_vnics() to get_connected_nics()

andy-maier:
Reason: The HMC API book is inconsistent in its use of VNIC vs. NIC. For example, the 'Get Connected VNICs of a Virtual Switch' HMC operation uses the term 'VNICs' for the same objects the 'Create NIC' HMC operation uses 'NIC' for. Therefore, we should use one term ('NIC') consistently in the zhmcclient API.

This PR is on top of PR #65.

Please review and merge (after merging PR #65).


leopoldjuergen:
+1

PR #75, opened 2016-09-15
Optimized dpm_enabled implementation for better performance.

andy-maier:
This speeds it up on an EC12 from 35..45 sec to 0.1 sec.

Please review.


leopoldjuergen:
+1

PR #76, opened 2016-09-15
Added cpcinfo tool and prop() method for resources.

andy-maier:
Please review.

Details:
- Added a script 'tools/cpcinfo' that displays information about CPCs.
- In support of that script, added a 'prop()' method for resources that allows specifying a default value in case the property does not exist.


leopoldjuergen:
+1

PR #77, opened 2016-09-16
Added more HBA/storage related methods.

andy-maier:
Details:
- Added `get_wwpns()` to `Cpc` class, which performs HMC operation 'Export WWPN List', and extended the unit test cases for it.
- Added `reassign_port()` to `Hba` class, which performs HMC operation 'Reassign Storage Adapter Port', and extended the unit test cases for it.
- Added 'FCP Port' as a term to the resource model (docs only).

Please review.


leopoldjuergen:
+1

PR #81, opened 2016-09-20
Several logistical cpcinfo fixes.

andy-maier:
Please review.

Details:
- Added tabulate and progressbar2 as prereqs in requirements.txt.
- Added tools/cpcinfo as a script in setup.cfg to get it installed into the PATH.
- Added a --version option to cpcinfo that displays its own/zhmcclient version and the Python version.


andy-maier:
+1.
Need to merge this quickly in order to address a user's problem with it.

PR #82, opened 2016-09-20
Fixes #10: Made unit tests consistent and split up unit tests and function tests.

### LINKED ISSUE #10: Unit Tests should be moved to 'unit' folder in tests folder
In the future we will not only have unit tests but also functional test and other tests. We should move the unit tests to 'tests' folder.

andy-maier:
Please review.

This PR contains both formal changes to the test cases for consistency, as well as the changes to reorganize the directory structure as proposed in issue #10.

Details:
- Moved unit tests to `tests/unit` and added `tests/function`.
- Made sure all unit tests import absolute_import and print_function from future.
- Made sure all test cases that can execute as programs, have a hashbang for Python, a check for invocation as a script that runs unittest.main(), and have permissions 755.
- Made sure all test cases that cannot execute as programs, have no hashbang, have no check for script invocation, and have permissions 644.
- Changed any imports from zhmcclient objects to be from zhmcclient namespace, instead of referencing the sub-namespaces.


andy-maier:
+1

preethipy:
Question: I understand it is mainly for python compatibility which if mentioned works with both 2.x and 3.x release. Do we have to bring in the __future_ import for print function even though we are no using it? 
https://wiki.python.org/moin/PortingPythonToPy3k there are other options to have compatible python support.

andy-maier:
Preethi, with "there are other options" do you mean the three options pointed out in that article:

* maintaining a python 2 base and using 2to3 to generate the python 3 code
* maintaining a python 3 base and using 3to2 to generate the python 2 code
* maintaining a code base that runs on both python 2 and 3

If so, we should go for the last of these options, IMO, and that's what the current code base does. The first two approaches do not cost less, and they have their own complexities (e.g. limitations of these tools).

With that last option, the `__future__` imports are sometimes needed to enable compatibility.

The difficulty with the `print_function` future is that even when it is not imported, python 2 has a `print()` function. What the `print_function` future does is to disable the `print` statement.

I found that sometimes in python 2 the `print()` function behaves differently when importing the `print_function` future compared to when not importing it. I did not investigate why that happens and what the differences were, but the ones I ran into were such that my code did not work. Maybe it was caused by some package overriding `print()` (providing that ability was one explicit goal for python to change from the `print` statement to the `print()` function). Too bad I did not take notes.

So I simply tried to be on the careful side here. If someone uses `print()` in the tests for debugging, then I wanted to make sure the same python 3 compatible `print()` is used.

However, that motivation may not be valid enough, and I don't have a strong opinion on that. Let me know if you think that the `print_function` import should be removed again.

andy-maier:
Rebased.

preethipy:
Was just curious about the print_function usage. I understand you are trying to be future proof :) the details given by you explains

PR #83, opened 2016-09-20
Added an install test to Travis CI and to Tox.

andy-maier:
Please review.
    
Details:
- In order to detect missing prereqs in non-development mode, added an install test to Travis CI that installs the package before installing the development dependencies.
- As part of that, changed the `make install` processing to install the package using `pip install .` (from `python setup.py install`), and added an import test to `make install`.
- For the Tox test, it was not easily possible to test installation without also having the development prereqs installed, but added the install test anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ python:
 install:
   - pip list
   - pip install --upgrade pip
+  - pip install .
+  - pip list
   - make develop
   - pip list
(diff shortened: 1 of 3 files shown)


andy-maier:
+1

andy-maier:
Merging this despite only one review, in order to build on it.

PR #88, opened 2016-09-21
Added HTTP method and URI to HTTPError information.

andy-maier:
This is in support of an HTTP error Elisabeth sees, and allows to find the operation in the HMC API book that documents the HTTP status code and reason code.

Please review.


preethipy:
+1

andy-maier:
+1

PR #90, opened 2016-09-23
Fixed HTTPError in dpm_enabled.

andy-maier:
Please review.

This issue went by undetected because the mocked HMC responses in the unit tests made the same mistake as the code that was tested. This shows the importance of a real test.

Details:
- In the optimized implementation of dpm_enabled, the http_status and reason properties of HTTPError were incorrectly tested against strings. Fixed that by testing against integers.
- The mocked HMC responses defined in the unit tests had the same problem and returned strings for http_status and reason, which is why the problem went in, undetected. Fixed that also by changing the mocked responses to use integers.


andy-maier:
+1

leopoldjuergen:
+1

PR #96, opened 2016-10-04
Improved ParseError and its use.

andy-maier:
This PR addresses the ParseError portion of issue #94.
The other parts will be addressed separately.

This PR is based on PR #95, so the first commit has already been reviewed as part of its review.

Please review and merge.

Details:
- Added line and column properties ot ParseError, which are parsed from the message string.
- In the low level HTTP methods in the Session class, making sure that ParseError gets raised if the returned JSON cannot be decoded.
- Extended the unit test cases for ParseError to test the additional line and col properties.
- Extended the unit test cases for Session to test some cases of invalid JSON being returned.

PR #95, opened 2016-10-04
Fixed change log to represent the already released v0.5.0.

andy-maier:
Please review.

This fixes the change log to reflect v0.5.0, and updates the how-to-release section to reflect the updating of teh change log better.

Because v0.5.0 is already released, I suggest to merge this into master.

Issue #94, opened 2016-10-04
Repetitive documentation of the raised exceptions


andy-maier:
Several issues:
1. We wrapper some exceptions into zhmcclient exceptions, but loose information. That is the case for:
   - AuthError - we create that for any HTTP status 403, and loose the details returned by the HMC.
   
   Status: **Done** via PR #97.
   - ConnectionError - while we include the message of the underlying exception in the message of the ConnectionError, we loose details that could be machine-processible such as the socket errno number etc.
     
     Status: **Done** via PR #97.
2. We do not raise the following exceptions at all, currently:
   - ParseError
     
     Status: **Done** via PR #96.
   - VersionError
     
     Need to revisit whether we want VersionError to be raised. Discussion with Juergen: We may have the situation that a zhmcclient function depends on a minimum HMC version or HMC APi version, so that the exception could be used in the future. We keep the exception defined, for now. However, looking at the exception, it should be improved to capture the expected vs. available version, and which type of version the error is about (HMC API vs HMC itself).

     Status: **Done** via PR #110.
3. The zhmcclient methods mostly repeat the zhmcclient exceptions in their Raises: definition. Can we simplify that.
   
   Status: **Open** See **discussion point**, below.
4. The zhmcclient Python properties can raise exceptions but do not document that at all currently.
   
   Status: **Done** Verified that all Python properties that possibly can raise exceptions, document that.

andy-maier:
Investigation of all Python properties shows that the ones that possibly can raise an exception all documented their exceptions in the docstring.

andy-maier:
**Discussion point:** Can we find a way how the repetition of exceptions raised by lower level functions in the docstring of higher level functions can be avoided, without impacting the usability of the documentation for the client.

andy-maier:
Adjusted the title of this issue to the remaining documentation issue.

andy-maier:
There does not seem to be an out of the box way to avoid repeating the exceptions. Closing this one.

PR #97, opened 2016-10-05
Improved the AuthError and ConnectionError exceptions and their use.

andy-maier:
This PR addresses the AuthError and ConnectionError portion of issue #94.
The other parts will be addressed separately.

This PR is based on PRs #95 and #96, so the first two commits have already been reviewed as part of their review.

Please review and merge.

Details:
- In `AuthError`, added a `details` property that is the underlying exception, and that is set to an `HTTPError` object in case the error is triggered by a failed HTTP response.
- In `ConnectionError`, added a `details` property that is the underlying exception, and that is set to a `RequestException` object in case the error is triggered by a failed requests method.
- Extended the unit testcases for `AuthError` and `ConnectionError` to test the new `details` property.
- Fixed a bug in the unit test cases for HTTPError, that caused the simple tests defined in `SimpleTestMixin` to be executed for the default exception class `Exception`, instead of `HttpError`. Changed the default exception class in `SimpleTestMixin` to `None`, in order to avoid such mistakes in the future.
- In the low level HTTP methods of the `Session` class, optimized the conversion of the response body into a Python dictionary (=JSON object) to avoid multiple conversions. In addition, separated that conversion onto a separate line for better debugging.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -32,6 +32,10 @@ Released: Not yet
 
 **Enhancements:**
 
+* Started raising a `ParseError` exception when the JSON payload in a HTTP
+  response cannot be parsed, and improved the definition of the ParseError
+  exception by adding line and column information.
+
 **Known Issues:**
 
 * See `list of open issues`_.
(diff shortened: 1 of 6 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #104, opened 2016-10-31
Session object can be created with session id

leopoldjuergen:
Details:
- A session object be created by a existing session id.
- Remove double assignment of _header and session_id.

### DIFF
--- a/zhmcclient/_session.py
+++ b/zhmcclient/_session.py
@@ -60,6 +60,35 @@ class Session(object):
         Creating a session object will not immediately cause a logon to be
         attempted; the logon is deferred until needed.
 
+        There are several alternatives for specifying the authentication
+        related parameters:
+
+        * `userid`/`password` only: The session is initially in a logged-off
+          state and subsequent operations that require logon will use the
+          specified userid and password to automatically log on. The returned
+          session-id will be stored in this session object. Subsequent
+          operations that require logon will use that session-id. Once the HMC
+          expires that session-id, subsequent operations that require logon
+          will cause a re-logon with the specified userid and password.
+
+        * `userid`/`password` and `session_id`: The specified session-id will
+          be stored in this session object, so that the session is initially in
+          a logged-on state. Subsequent operations that require logon will use
+          that session-id. Once the HMC expires that session-id, subsequent
+          operations that require logon will cause a re-logon with the
+          specified userid/password.
+
+        * `session_id` only: The specified session-id will be stored in this
+          session object, so that the session is initially in a logged-on
+          state. Subsequent operations that require logon will use the stored
+          session-id. Once the HMC expires the session-id, subsequent
+          operations that require logon will cause an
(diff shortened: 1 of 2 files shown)


leopoldjuergen:
+1

andy-maier:
Please describe why this is needed.

zhmcclient/_session.py
```diff
@@ -75,6 +75,11 @@ def __init__(self, host, userid=None, password=None):
           password (:term:`string`):
             Password of the HMC user to be used.
 
+          session_id (:term:`string`):
+            An opaque string that provides a cryptographically
+            strong identifier of the API session (known as a session id)
+            under which this request is executed.
+
```
andy-maier: Some comments on the description of the new `session_id` argument of `Session.__init__()`:

1. On "An opaque string that provides a cryptographically strong identifier of the API session (known as a session id)": I suggest to use the same wording we use for the `session_id` property: "session ID".

2. The description does not make it clear enough that providing a session ID puts the session into the logged-on state.

3. It is not clear how the `session_id` and the `userid` / `password` arguments play together.
 
I suggest the following description for the `session_id` argument, to address these issues:

```
    A session ID returned by the HMC that is to be used for this session.
    Supplying a session ID puts the session into a logged-on state.

    The `userid` and `password` arguments may be omitted if a session ID
    is supplied. If they are omitted, expiration of the session ID causes an exception
    to be raised by subsequent operations, instead of performing an automatic
    re-logon.
```

zhmcclient/_session.py
```diff
@@ -85,11 +90,11 @@ def __init__(self, host, userid=None, password=None):
             host=self._host,
             port=_HMC_PORT)
         self._headers = _STD_HEADERS  # dict with standard HTTP headers
-        self._session_id = None  # HMC session ID
+        self._session_id = session_id  # HMC session ID
+        if self._session_id is not None:
+            self._headers['X-API-Session'] = self._session_id
         self._session = None  # requests.Session() object
```
andy-maier: The `Session._do_logon()` method sets the `_session` instance variable. I think we should also do this if a session ID was supplied. After all, we want to behave the same way as if a logon had been performed.

andy-maier:
-1, see my two comments.

zhmcclient/_session.py
andy-maier: DONE: Addressed my comment in an update to this PR (differently that proposed in my comment, above).

zhmcclient/_session.py
andy-maier: DONE: In an update to this PR.

andy-maier:
Commit b9d902c addresses my earlier comments on this PR.
+1 now from me.
Please review.

andy-maier:
Rebased to master, in order to pick up the fix for the flake8 issues.

andy-maier:
+1

leopoldjuergen:
+1

PR #107, opened 2016-11-04
Added time statistics entry for overall async operations.

andy-maier:
So far,  the time statistics only records the actual async POST operation, i.e. the start of what happens in the background, and separately, all the GET operations that poll for job completion. This makes it hard to figure out what the elapsed time is for the entire operation.

This PR improves that by adding an entry in the time statistics for the overall operation consisting of the sequence of async POST plus all the GETs for polling.

Please review and merge.


andy-maier:
+1

PR #109, opened 2016-11-17
Fixed new Flake8 issue E305.

andy-maier:
This PR fixes the new E305 issue reported by the new version of flake8 that is used in the Travis CI.

This PR needs to be merged quickly, because all other PRs will also run into this issue.


leopoldjuergen:
+1

PR #119, opened 2016-12-07
Fixed incorrect package version for pypi upload.

andy-maier:
Please review and merge.

Details:
- Probably, the bug that an incorrect version was uploaded to Pypi was caused by having that version installed as a package in the virtual Python environment.
- Fixed that bug in the Makefile, by:
  - Adding an 'uninstall' target that uninstalls the Python package from the current Python environment.
  - Invoking that 'uninstall' target as part of the 'clobber' and 'upload' targets.
  - As an additional safety net, added a check to the 'upload' target that rejects the attempt to upload development versions.
  - As a second additional safety net, added a confirmation prompt in the 'upload' target that asks the user for confirmation.
- Updated the Pypi upload steps in the RTD documentation to state that a user confirmation is needed for 'make upload'.


leopoldjuergen:
+1

PR #123, opened 2016-12-08
Upgraded click-spinner to v0.1.5 and simplified its use

andy-maier:
This PR removes the pinning of click-spinner to 0.1.3 and instead requires >=0.1.5, and at the same time it exploits the new abilities in 0.1.5 (ability to pause via stop/start, and automatic suppression when not on a tty).

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -20,18 +20,16 @@ Change log
 ----------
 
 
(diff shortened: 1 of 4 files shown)


leopoldjuergen:
+1

PR #124, opened 2016-12-15
Added support in CLI for remaining cmds; client improvements.

andy-maier:
This PR is pretty much completely tested and ready for serious use.

zhmcclient library changes:
- Added support in the `BaseResource` class for optimized resource access by name. This includes:
  - an instance variable `_uris` that is a cached mapping of resource names to resource URIs
  - a new `find_by_name()` method that is automatically used by `findall()` and `find()` when called with 'name' as the single filter, but can also be called by the user.
  - a `flush()` method that flushes the cached name-to-URI mapping. This is only needed when renaming resources.
- Added two required parameters to the init method of class BaseResource. This method is not part of the external API, so this is not an incompatible change.

zhmc CLI changes:
- Added support for all missing resource types for DPM mode.
- Added and changed options for creating and updating resources for already supported resource types. Resources referenced in options are now consistently referenced by resource name.
- The resource name is now always part of the output.
- Added options for showing additional properties. The URI is now only shown when its option is used.
- In table output, resources are now sorted  by resource name.
- In table output, the columns now have a specific ordering.

### DIFF
--- a/tests/unit/test_hba.py
+++ b/tests/unit/test_hba.py
@@ -236,16 +236,16 @@ class HbaTests(unittest.TestCase):
         """
         hba_mgr = self.partition.hbas
         hba_uri = '/api/partitions/fake-part-id-1/hbas/fake-hba-id-1'
-        hba = Hba(hba_mgr, uri=hba_uri, properties={})
+        hba = Hba(hba_mgr, hba_uri)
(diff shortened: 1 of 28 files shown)


leopoldjuergen:
+1

PR #130, opened 2016-12-27
Add example for CLI zhmc

leopoldjuergen:
Add example for CLI zhmc
    
Details:
- Example implements create, delete, updates
  and list of a partition using the CLI.
- Example implements check of DPM mode.
- Example file provided for runtime configuration
  information stored in zhmcrc.sh file.
- zhmcrc.sh added to gitignore file to
  avoid accidentally release runtime information
  to repository.

### DIFF
--- a/examples/cli/crud_partition.sh
+++ b/examples/cli/crud_partition.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+echo "Example script demonstrating zhmc CLI partition commands."
+
+eval $(zhmc -h $HMC_HOST -u $HMC_USER session create)
+zhmc cpc list
+out=$(zhmc -o json cpc show $CPC_NAME)
+dpm=$(echo "$out" | python -c "import json,sys; obj=json.load(sys.stdin); print(obj['dpm-enabled'])")
+if [ $dpm != "True" ]
+then
+   echo "CPC $CPC_NAME is not in DPM mode."
+   eval $(zhmc session delete)
+   exit 2
+else
+   echo "CPC $CPC_NAME is in DPM mode."
+fi
+zhmc partition list $CPC_NAME
+zhmc partition create $CPC_NAME \
+   --name $PART_NAME \
+   --description "This is a partition created by zhmc." \
+   --cp-processors 1 \
+   --initial-memory 4096 \
+   --maximum-memory 4096 \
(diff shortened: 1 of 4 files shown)


examples/cli/crud_partition.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+echo "CRUD partition script."
```
andy-maier: I suggest to change this echo to something like "Example script demonstrating zhmc CLI partition commands."
Just in case users are not familiar with the acronym "CRUD".

examples/cli/crud_partition.sh
```diff
@@ -0,0 +1,16 @@
...
+HMC_HOST="Put here the HMC host"
+HMC_USER="Put here HMC user"
+CPC_NAME="Put here the name of the CPC"
+PART_NAME="TESTPARTITION"
+
+eval $(zhmc -h $HMC_HOST -u $HMC_USER session create)
+zhmc cpc list
+zhmc partition list $CPC_NAME
```
andy-maier: The script depends on the CPC being in DPM mode (which is fine).
It seems to me that we should demonstrate in the script how to check for DPM mode.

For example by parsing the output of partition show in JSON (not tested):
```
out=$(zhmc -o json cpc show $CPC_NAME)
dpm=$(echo "$out" |python -c 'import json,sys; obj=json.load(sys.stdin); print obj["dpm-enabled"]')
if [[ "$dpm" != "True" ]]; then
  echo "CPC $CPC_NAME is not in DPM mode."
  exit 2
fi
```

andy-maier:
-1, see my (minor) comments.

andy-maier:
One more comment: Please add an entry for this in the change history.

examples/cli/crud_partition.sh
leopoldjuergen: Yes, it makes sense to change this.

examples/cli/crud_partition.sh
leopoldjuergen: Good idea :-) I like the jq tool but I think the python solution is good enough and I agree it avoids a new dependency.

leopoldjuergen:
+1

andy-maier:
+1

PR #134, opened 2016-12-28
Add Metrics support to the API

leopoldjuergen:
Details:
- Add MetricsContext and MetricsContextManager class.
- Add Metrics resource to documentation.
- Add metrics example.

### DIFF
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -209,3 +209,27 @@ Virtual Switches
 .. autoclass:: zhmcclient.VirtualSwitch
    :members:
    :special-members: __str__
+
+
+.. _`Metrics`:
+
+Metrics
+-------
+
+.. automodule:: zhmcclient._metrics
+
+.. autoclass:: zhmcclient.MetricsContextManager
+   :members:
+   :special-members: __str__
+
+.. autoclass:: zhmcclient.MetricsContext
+   :members:
+   :special-members: __str__
+
+.. autoclass:: zhmcclient.CollectedMetrics
+   :members:
+   :special-members: __str__
+
+.. autoclass:: zhmcclient.Metrics
+   :members:
+   :special-members: __str__
--- a/docs/resources.rst
+++ b/docs/resources.rst
@@ -209,27 +209,3 @@ Virtual Switches
 .. autoclass:: zhmcclient.VirtualSwitch
    :members:
    :special-members: __str__
-
-
-.. _`Metrics`:
-
-Metrics
--------
-
-.. automodule:: zhmcclient._metrics
-
-.. autoclass:: zhmcclient.MetricsContextManager
-   :members:
-   :special-members: __str__
-
-.. autoclass:: zhmcclient.MetricsContext
-   :members:
-   :special-members: __str__
-
-.. autoclass:: zhmcclient.CollectedMetrics
-   :members:
-   :special-members: __str__
-
-.. autoclass:: zhmcclient.Metrics
-   :members:
-   :special-members: __str__
--- a/zhmcclient/_client.py
+++ b/zhmcclient/_client.py
@@ -67,8 +67,8 @@ class Client(object):
     def metrics_contexts(self):
         """
         :class:`~zhmcclient.MetricsContextManager`:
-          Manager object for the MetrcsContexts in scope of this client
-          (i.e. in scope of its HMC).
+          Manager object for the :term:`Metrics Contexts <Metrics Context>` in
+          scope of this client (i.e. in scope of its HMC).
         """
         return self._metrics_contexts
 
--- a/zhmcclient/_client.py
+++ b/zhmcclient/_client.py
@@ -19,6 +19,7 @@ Client class: A client to an HMC.
 from __future__ import absolute_import
 
 from ._cpc import CpcManager
+from ._metrics import MetricsContextManager
 from ._logging import get_logger, logged_api_call
 
 __all__ = ['Client']
@@ -42,6 +43,7 @@ class Client(object):
         """
         self._session = session
         self._cpcs = CpcManager(self)
+        self._metrics_contexts = MetricsContextManager(self)
         self._api_version = None
 
     @property
@@ -60,6 +62,15 @@ class Client(object):
           of its HMC).
         """
         return self._cpcs
+
+    @property
+    def metrics_contexts(self):
+        """
+        :class:`~zhmcclient.MetricsContextManager`:
+          Manager object for the MetrcsContexts in scope of this client
+          (i.e. in scope of its HMC).
+        """
+        return self._metrics_contexts
 
     @logged_api_call
     def version_info(self):
--- a/zhmcclient/_metrics.py
+++ b/zhmcclient/_metrics.py
@@ -1,0 +1,390 @@
+# Copyright 2016 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A Metrics Service provides a mechanism to retrieve performance
+metric data for resources of a physical z Systems or LinuxONE computer.
+A structure called a :term:`Metrics Context` is associated with
+any metrics retrieval, and that structure includes metric group names,
+individual metric field names, and the associated individual
+metric data types.
+"""
+
+from __future__ import absolute_import
+
+from ._manager import BaseManager
+from ._resource import BaseResource
+from ._logging import get_logger, logged_api_call
+from ._exceptions import NotFound
+
+__all__ = ['MetricsContextManager', 'MetricsContext', 'Metrics',
+           'CollectedMetrics']
+
+LOG = get_logger(__name__)
+
+
+class MetricsContextManager(BaseManager):
+    """
+    Manager providing access to the :term:`Metric sContexts <Metrics Context>`
+    exposed by the HMC this client is connected to.
+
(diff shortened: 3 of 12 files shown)


zhmcclient/_client.py
```diff
@@ -61,6 +63,15 @@ def cpcs(self):
         """
         return self._cpcs
 
+    @property
+    def metrics_contexts(self):
+        """
+        :class:`~zhmcclient.MetricsContextManager`:
+          Manager object for the MetrcsContexts in scope of this client
```
andy-maier: * Typo "MetrcsContexts"
* In the text, we should either talk about "Metrics Contexts" where both words are a link to the same-named glossary term, or about "MetricsContext objects" where the first word is a link to the same-named class.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+
+__all__ = ['MetricsContextManager', 'MetricsContext', 'Metrics',
+           'CollectedMetrics']
+
+LOG = get_logger(__name__)
+
+
+class MetricsContextManager(BaseManager):
+    """
+    Manager providing access to the :term:`Metric sContexts <Metrics Context>`
```
andy-maier: * Typo "Metric sContexts"

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+            uri_prop='object-uri',
+            name_prop='',
+            query_props=query_props)
+
+        self._session = client.session
+        self._metrics_contexts = []
+        self._client = client
+
+    @logged_api_call
+    def list(self, full_properties=False):
```
andy-maier: The current HMC REST API does not allow listing the existing Metrics Context objects.
Suggest to have a `list()` function that raises `NotImplemented` and that documents that metrics context objects cannot be listed.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        List the MetricsContext exposed by the HMC this client is connected to.
+
+        Parameters:
+
+          full_properties (bool):
+            Controls whether the full set of resource properties should be
+            retrieved, vs. only the short set as returned by the list
+            operation.
```
andy-maier: This parameter does not influence anything. I suggest having it for compatibility, but describing that it will be ignored.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+
+        self._session = client.session
+        self._metrics_contexts = []
+        self._client = client
+
+    @logged_api_call
+    def list(self, full_properties=False):
+        """
+        List the MetricsContext exposed by the HMC this client is connected to.
+
```
andy-maier: I first thought that having a `list()` function does not work because the HMC does not allow listing them, but at a second thought this seems to work.

However: We need to better point this out in the description of `list()`, e.g.:

    List the :term:`Metrics Context`s that have been created through this session.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+__all__ = ['MetricsContextManager', 'MetricsContext', 'Metrics',
+           'CollectedMetrics']
+
+LOG = get_logger(__name__)
+
+
+class MetricsContextManager(BaseManager):
+    """
+    Manager providing access to the :term:`Metric sContexts <Metrics Context>`
+    exposed by the HMC this client is connected to.
```
andy-maier: "exposed by the HMC this client is connected to." is not really correct, because the HMC does not expose them at all (they cannot be listed). Change to, e.g:

    that have been created through this client.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        if not isinstance(manager, MetricsContextManager):
+            raise AssertionError(
+                "MetricsContext init: Expected manager type %s, got %s" %
+                (MetricsContextManager, type(manager)))
+        super(MetricsContext, self).__init__(manager, uri, name, properties)
+
+    @logged_api_call
+    def get_metrics(self):
+        """
+        Start this CPC, using the HMC operation "Start CPC".
```
andy-maier: Wrong description.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Raises:
+
+          :exc:`~zhmcclient.HTTPError`
+          :exc:`~zhmcclient.ParseError`
+          :exc:`~zhmcclient.AuthError`
+          :exc:`~zhmcclient.ConnectionError`
+        """
+        result = self.manager.session.get(self.uri)
+        return result
```
andy-maier: We should not hand back the result as a JSON object, but as a proper Python object (I would expect a list of `MetricsGroup` objects, because a metrics context stands for a list of metrics groups)

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+    @logged_api_call
+    def get_metrics(self):
+        """
+        Start this CPC, using the HMC operation "Start CPC".
+
+        Returns:
+
+          :term:`json object`:
+
+            If `wait_for_completion` is `True`, returns None.
```
andy-maier: Wrong description.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        Delete this MetricsContext.
+
+        Raises:
+
+          :exc:`~zhmcclient.HTTPError`
+          :exc:`~zhmcclient.ParseError`
+          :exc:`~zhmcclient.AuthError`
+          :exc:`~zhmcclient.ConnectionError`
+        """
+        self.manager.list().remove(self)
```
andy-maier: What if we make `list()` return a copy?

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        self.manager.session.delete(self.uri)
```
andy-maier: What if this fails? Then we have already removed it from the instance variable. The order should be switched around, IMO.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class CollectedMetrics(object):
+
```
andy-maier: Description missing.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+    def __init__(self, metrics_context, rawdata):
+        """
```
andy-maier: Clarify for the reader of the documentation whether objects of this class can be created by the user vs. are being reated by some function. In the latter case, use the approach we use for manager and resource objects on how to omit the constructor from the documentation.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+                        m = Metrics(self._client, metrics_group, uri,
+                                    epoch_timestamp, metrics)
+                        metrics_list.append(m)
+                        state = 4
+            self._metrics = metrics_list
+            return self._metrics
+
+
+class Metrics(object):
+
```
andy-maier: Description missing.

zhmcclient/_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+            metric_infos = metric_group_info['metric-infos']
+            metric_names = list()
+            for metric_info in metric_infos:
+                metric_names.append(metric_info['metric-name'])
+            metric_groups[metric_group_info['group-name']] = metric_names
+        self._metrics_groups = metric_groups
+
+    @property
+    def metrics(self):
+        if self._metrics:
```
andy-maier: Description missing.

docs/resources.rst
```diff
@@ -177,3 +177,27 @@ Virtual Switches
 .. autoclass:: zhmcclient.VirtualSwitch
    :members:
    :special-members: __str__
+
+
```
andy-maier: The "Metrics" section should not be in the "Resources" chapter of the documentation, because it is not really a resource. Either we create a "Services" chapter, or we add it to the "General features" chapter.

andy-maier:
Comments:
* See my comment 2 on PR #220 abut adding the metrics unit.
* See my inline comments.
* I have made some inline comments on the `CollectedMetrics` and `Metrics` classes, but I suggest that we rethink the structure of the returned metrics data, considering this:
  - The return value from `get_metrics()` I'd expect to be a list of `MetricsGroup` objects, which then break down into the lower units until we arrive at single metrics.
  - The metric descriptions should probably not be flattened into the single metrics. That would blow up the metrics too much, IMO. It would also make it more difficult to get at the metric definition independently of the metric values.

docs/resources.rst
andy-maier: DONE: Moved to "General features".

zhmcclient/_client.py
andy-maier: DONE

zhmcclient/_metrics.py
andy-maier: DONE

zhmcclient/_metrics.py
andy-maier: DONE: Left unchanged. Even though they are created by the client, afterwards they are exposed by the HMC.

zhmcclient/_metrics.py
andy-maier: DONE: left unchanged.

zhmcclient/_metrics.py
andy-maier: DONE as propose.d

zhmcclient/_metrics.py
andy-maier: DONE

zhmcclient/_metrics.py
andy-maier: DONE

zhmcclient/_metrics.py
andy-maier: For now, added an according TODO.

zhmcclient/_metrics.py
andy-maier: DONE. Changed to use the attribute directly.

zhmcclient/_metrics.py
andy-maier: DONE. Switched order around.

andy-maier:
Rebased and addressed my comments in a second commit.

andy-maier:
In the last commit, I have reworked the way the MetricsResponse is processed.

PR #136, opened 2017-01-03
Added support for server-side resource filtering.

andy-maier:
This PR adds support for filtering properties on the server side via filter query parameters. This is done transparently, for those properties that are supported. The rest of the properties is still filtered on the client side.

Please review and merge.

There are two incompatibilities:
* The support for server-side filtering caused an incompatibility for the `find()` and `findall()` methods: For String typed resource properties, the provided filter string is now interpreted as a regular expression that is matched against the actual property value, whereby previously it was matched by exact string comparison.
* The parameter signatures of the `__init__()` methods of `BaseResource` and `BaseManager` have changed incompatibly. These methods have always been  considered internal to the package. They are now explicitly stated to be internal and their parameters are no longer documented. If users have made themselves dependent on these parameters (e.g. by writing a mock layer), they will need to adjust to the new parameter signature. See the code for details.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -27,11 +27,33 @@ Released: not yet
 
 **Incompatible changes:**
 
+* The support for server-side filtering caused an incompatibility for the
+  `find()` and `findall()` methods: For String typed resource properties,
+  the provided filter string is now interpreted as a regular expression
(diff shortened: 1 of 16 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #138, opened 2017-01-03
Fixes #137: Accessing properties['name'] may raise KeyError.

### LINKED ISSUE #137: find() or findall() with 'name' fails with KeyError
### Actual behavior

Invoking any `find()` or `findall()` method with just the `name` property raises `KeyError` when accessing `obj.properties['name']`. This was introduced with the URI-to-name cache recently.

This causes example2.py to fail that way, and probably most others, too.

### Expected behavior

This should work.

### Execution environment

* zhmcclient version: 0.8.0
* Operating system (type+version): RHEL 7.2

andy-maier:
Please review and merge.

Details from the commit message:
* This bug was introduced recently with the URI-to-name caching, which now causes the 'properties' attribute to be empty where it was not empty before.
* Fixed by maintaining the 'name' property in the 'properties' attribute whenever possible (i.e. available). This does not solve all situations, therefore updated the documentation of the 'properties' attribute to make users aware, and to recommend use of 'get_property()' or 'name' or 'uri', for accessing specific resource properties.
* Also, added assertions in `BaseResource.__init__()` that ensure that the uri and the name are consistent, if provided both ways (via direct argument and via the 'properties' dict).
* Updated the examples accordingly. They all runs again, now.
* In example2, removed the storing of the job in a 'status' variable because it is mis-named and was not used anyway.
* In example 4, changed the variables for storing the async job and its URI more logical, and made the prints more logical.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,6 +31,13 @@ Released: not yet
 
 **Bug fixes:**
 
+* Fixed a bug where accessing the 'name' property via the `properties`
+  attribute caused `KeyError` to be raised (issue #137). Note that there
+  is now a recommendation to use `get_property()` or the `name` or `uri`
+  attributes for accessing specific properties. The `properties` attribute
+  should only be used for iterating over the currently present resource
+  properties, but not for expecting particular properties.
+
 **Enhancements:**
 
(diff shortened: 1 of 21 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #143, opened 2017-01-10
[WIP] Mocking support for zhmcclient.

andy-maier:
Initial stab, not yet done.

Besides the code, have a look at the example testcase (`tests/unit/zhmcclient_mock/test_example.py`) and at the new chapter "Mock support" in the generated documentation (`make builddoc`, to generate it).

The new example unit test fails because the implementation is not complete yet.


andy-maier:
This PR is abandoned because it will be split into two PRs.

PR #144, opened 2017-01-11
Design for mocking support for zhmcclient

andy-maier:
Please review.

The sibling PR #143 provides a first implementation of the design.

### DIFF
--- a/design/fake-client.rst
+++ b/design/fake-client.rst
@@ -1,0 +1,153 @@
+.. Copyright 2016 IBM Corp. All Rights Reserved.
+..
+.. Licensed under the Apache License, Version 2.0 (the "License");
+.. you may not use this file except in compliance with the License.
+.. You may obtain a copy of the License at
+..
+..    http://www.apache.org/licenses/LICENSE-2.0
+..
+.. Unless required by applicable law or agreed to in writing, software
+.. distributed under the License is distributed on an "AS IS" BASIS,
+.. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+Design for fake client
+======================
+
+This document describes requirements and possible design options for a fake
+zhmcclient.
+
+Requirements
+------------
+
+* Must avoid the need for a real HMC and still allow usage of the zhmcclient
+  API.
+* Both read-only and read-write operations must be supported.
+* Read-write operations must reflect their changes properly (i.e. they must be
+  visible in corresponding read operations).
+* The resource state of the faked HMC must be preloaded (e.g. for resources
+  that cannot be created by clients), and its resource descriptions must be
+  easy to use (i.e. not in terms of REST URIs.
+* It must be possible to define testcases with error injection.
+
+It seems that with these requirements, we can assume that there is a need for
+a faked HMC that can represent its resources and properties as Python objects,
+so that changes can be applied consistently.
+
+Design options
+--------------
+
+There are two basic design options for the fake client:
+
+* Complete re-implementation of the zhmcclient API, that behaves the same
+  way as the zhmcclient package, but consists of a different classes that
+  implement a faked HMC.
+* Usage of the real zhmcclient package, and replacing its Session object
+  with a fake session that implements a faked HMC.
+
+The first design option is probably less effort, but it has the drawback that
+subtle changes in the API may be forgotten. One example for that is the
+recent introduction of an optimization for findall() by name, which now uses
+a cached mapping of names to URIs, but had the subtle externally visible
+change that the resource objects returned by findall(name) now only have
+a very minimal set of properties, and no longer those that are returned by
+list() which was the earlier implementation. For example, the "status"
+property is no longer set in that case. The second design would automatically
+represent this example of a change.
+
+Design
+------
+
+In the following description, the second design option is used.
+
+There is a fake session object that the user of the zhmcclient package
+will use instead of zhmcclient.Session. The fake session object contains a
+faked HMC that can represent all resources relevant for the zhmcclient.
+There fake session object implements the same API as zhmcclient.Session, so
+that the rest of the zhmcclient classes does not need to be changed.
+The get(), post() and delete() methods of the fake session class will
+be redirected to operate on the faked HMC.
+
+The fake session object will have the ability to be "loaded" with an
+initial state of resources. The resource definitions can be provided as
+a dictionary of nested dictionaries, following this example::
+
+    resources = {
+        'cpcs': [  # name of manager attribute for this resource
+            {
+                'properties': {
+                    # object-id and object-uri are auto-generated
+                    'name': 'cpc_1',
+                    . . .  # more properties
+                },
+                'adapters': [
+                    {
+                        'properties': {
+                            # object-id and object-uri are auto-generated
+                            'name': 'ad_1',
+                            . . .  # more properties
+                        },
+                        'ports': [
+                            {
+                                'properties': {
+                                    # element-id and element-uri are
+                                    # auto-generated
+                                    'name': 'port_1',
+                                    . . .  # more properties
+                                }
+                            },
+                            . . .  # more Ports
+                        ],
+                    },
+                    . . .  # more Adapters
+                ],
+                . . .  # more CPC child resources
+            },
+            . . .  # more CPCs
+        ]
+    }
+
+The properties for object ID ("object-id" or "element-id") and URI
+("object-uri" or "element-uri") are auto-generated, if not provided by
+the user. All other properties will only exist as provided by the user.
+There is no checking for invalid properties.
+
+For all operations against the faked HMC, a successful operation is
+implemented by default. Only if a testcase needs error injection, the user
+needs to deal with setting up the behavior of faked HMC operations.
+
+To support this, the fake session object will have the ability to specify
+non-standard behavior of HMC operations, at the level of the HTTP interactions,
+following this example::
+
+    operations = [
+        {
+            'method': 'get',           # used for matching the request
+            'uri': '/api/version',     # used for matching the request
+            'request_body': None,      # used for matching the request
+            'response_status': 200,    # desired HTTP status code
+            'response_body': {         # desired response body
+                'api-minor-version': '1'
+            }
+        },
+        {
+            'method': 'get',           # used for matching the request
+            'uri': '/api/cpcs',        # used for matching the request
+            'request_body': None,      # used for matching the request
+            'response_status': 400,    # desired HTTP status code
+            'response_error': {        # desired error info
+                'reason': 25,          # desired HMC reason code
+                'message': 'bla',      # desired HMC message text
+        }
+    ]
+
+This information is stored in the fake session and used in matching
+HTTP requests instead of the standard, successful implementations.
+
+It allows specifying successful operations with responses deviating from the
+standard responses (see the first list item in the example above) as well as
+error responses (see the second list item). Because the zhmcclient only
+evaluates HTTP status code, HMC reason code and the message text, these are the
+only attributes that can be specified (to keep it simple).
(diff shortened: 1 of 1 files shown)


design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+There fake session object implements the same API as zhmcclient.Session, so
+that the rest of the zhmcclient classes does not need to be changed.
+The get(), post() and delete() methods of the fake session class will
+be redirected to operate on the faked HMC.
+
+The fake session object will have the ability to be "loaded" with an
+initial state of resources. The resource definitions can be provided as
+a dictionary of nested dictionaries, following this example::
+
+    resources = {
```
scheuran: I like that format! It's easy to use!

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+
+For all operations against the faked HMC, a successful operation is
+implemented by default. Only if a testcase needs error injection, the user
+needs to deal with setting up the behavior of faked HMC operations.
+
+To support this, the fake session object will have the ability to specify
+non-standard behavior of HMC operations, at the level of the HTTP interactions,
+following this example::
+
+    operations = [
```
scheuran: As a user, I would not go this way. I would rather have a look at the exceptions that are thrown by a method, would mock the appropriate operation with a side_effect which throws this exception. I don't want to deal with APIs, error codes and so on.

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+                . . .  # more CPC child resources
+            },
+            . . .  # more CPCs
+        ]
+    }
+
+The properties for object ID ("object-id" or "element-id") and URI
+("object-uri" or "element-uri") are auto-generated, if not provided by
+the user. All other properties will only exist as provided by the user.
+There is no checking for invalid properties.
```
scheuran: it would be benefitial to have that but I see that this would be a huge effort...

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+* Read-write operations must reflect their changes properly (i.e. they must be
+  visible in corresponding read operations).
+* The resource state of the faked HMC must be preloaded (e.g. for resources
+  that cannot be created by clients), and its resource descriptions must be
+  easy to use (i.e. not in terms of REST URIs.
+* It must be possible to define testcases with error injection.
+
+It seems that with these requirements, we can assume that there is a need for
+a faked HMC that can represent its resources and properties as Python objects,
+so that changes can be applied consistently.
```
markuszoeller: Do you want to say here that this faked HMC is stateful? IOW, is this (pseudocode example below) possible?
```
assert 0 == cpc.partitions.list()
partition = cpc.partitions.create()
assert 1 == cpc.partitions.list()
partition.delete()
assert 0 == cpc.partitions.list()
```

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+
+* Must avoid the need for a real HMC and still allow usage of the zhmcclient
+  API.
+* Both read-only and read-write operations must be supported.
+* Read-write operations must reflect their changes properly (i.e. they must be
+  visible in corresponding read operations).
+* The resource state of the faked HMC must be preloaded (e.g. for resources
+  that cannot be created by clients), and its resource descriptions must be
+  easy to use (i.e. not in terms of REST URIs.
+* It must be possible to define testcases with error injection.
```
markuszoeller: I'm not sure if this is even necessary. The error can be mocked on the library user side with standard means of a decent mocking framework (`side_effect` of the `mock` package, for example): https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+
```
markuszoeller: Although it is probably implied, it would be great to have here the explicit statement that the solution, however it might look like, must be usable in *unit tests* of library users. This means the solution needs to perform fast and doesn't rely on external things, like a small HTTP server or similar.

design/fake-client.rst
andy-maier: Checking for invalid properties would require to know the names of all properties f all resource types. Right now, the zhmcclient code and the prototype code for this fake client have to know only a very small number of properties (object-id, object-uri, name). Plus, the set of properties depends on the API version, and probably some even depend on the CPC generation. Trying to avoid that knowledge about all properties was really the reason for this statement.

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+    resources = {
+        'cpcs': [  # name of manager attribute for this resource
+            {
+                'properties': {
+                    # object-id and object-uri are auto-generated
+                    'name': 'cpc_1',
+                    . . .  # more properties
+                },
+                'adapters': [
```
markuszoeller: How is the structure of this pre-loaded dict defined? For example `adapters`, does the name of the key have to be the exact same as a resource name? I don't see the benefit of using a dict, to be honest. How about setting up the state like this:
```
hmc = faked_hmc()
cpc = Cpc(name='cpc1')
adapter = Adapter(name='ad_1')
cpc.adapters.add(adapter)
hmc.cpcs.add(cpc1)

assert 'cpc1' == hmc.cpcs.list()[0].name
```
I think that's more re-usable than a complex dict. For example, let it be 5 CPCs which are managed by one HMC. The dict here gets really big with a lot of redundancy.

design/fake-client.rst
andy-maier: Do we have a requirement to be able to specify that a particular operation/method succeeds a few times and then fails?

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+Design for fake client
+======================
+
+This document describes requirements and possible design options for a fake
+zhmcclient.
+
+Requirements
+------------
+
+* Must avoid the need for a real HMC and still allow usage of the zhmcclient
```
andy-maier: with simulated HMC, you mean OCSE? Or the faked HMC we would set up?

design/fake-client.rst
andy-maier: agreed

design/fake-client.rst
andy-maier: yes. the state is maintained, and your pseudocode would pass

andy-maier: PS: that's the part I have prototyped already.

design/fake-client.rst
andy-maier: In realistic scenarios, this is going to specify quite a number of resources. For example, anything that cannot be created through the zhmcclient API, e.g. Adapters, Ports, any pre-existing partitions andtheir child resources, anything for classic mode. The lower level interface you pointed out is also available (it is what the function interpreting the dictionary issues). So I guess we would have both possibilities.

design/fake-client.rst
markuszoeller: I mean OCSE, yes.

design/fake-client.rst
markuszoeller: I guess that's outside of the scope of unit tests from a zhmcclient library user.

andy-maier: it is not. consider create partition, which has a limit.

design/fake-client.rst
scheuran: I prefer the dict approach - as I can easily see how my hmc looks like at first sight. with the programmed approach, it's hard to see and understand the nesting at first sight. I'm fine if both are available..

design/fake-client.rst
scheuran: I guess I would mock hat with python means again... I guess this is similar to the input parameter checking.
It would be really great to have that kind of limits also available in the faked version. Otherwise I have the same risk of doing things wrong like with my faked approach.

design/fake-client.rst
andy-maier: DONE: Added a brief reasoning why this is a huge effort.

design/fake-client.rst
andy-maier: DONE: Switched from the special operations approach to recommending side_effects. Demoted the special operations approach to a possible future extension, if needed.
DONE: Added need to represent limits for certain resources.

design/fake-client.rst
andy-maier: DONE: Replaced "real HMC" with "actual HMC" to be more neutral.

design/fake-client.rst
andy-maier: I agree that using side_effects should be promoted, but that does not mean that the requirement is invalid. Let me know if you think the requirement should be reworded.

design/fake-client.rst
andy-maier: DONE: Added the requirements:
* The solution must be usable in unit tests of zhmcclient users.
* The solution needs to perform fast, because it is used in a unit test.
* The solution should not rely on external componentry such as a small HTTP server.

design/fake-client.rst
andy-maier: DONE: Nothing to do.

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+
+Design
+------
+
+In the following description, the second design option is used.
+
+There is a fake session object that the user of the zhmcclient package
+will use instead of zhmcclient.Session. The fake session object contains a
+faked HMC that can represent all resources relevant for the zhmcclient.
+There fake session object implements the same API as zhmcclient.Session, so
```
andy-maier: DONE

design/fake-client.rst
```diff
@@ -0,0 +1,153 @@
...
+that the rest of the zhmcclient classes does not need to be changed.
```
andy-maier: DONE

design/fake-client.rst
andy-maier: DONE: To answer the uncertainty about the dictionary key names, a section was added describing that in more detail. On the general approach: The fake HMC already has methods to add resources, pretty much in the way Markus proposed it. Added a statement to the design spec that such direct methods can also be used as an alternative to the dictionary approach, to populate / manipulate the fake HMC resource state.

design/fake-client.rst
```diff
@@ -63,16 +67,22 @@ Design
...
 be redirected to operate on the faked HMC.
 
-The fake session object will have the ability to be "loaded" with an
-initial state of resources. The resource definitions can be provided as
-a dictionary of nested dictionaries, following this example::
+The fake session object will have the ability to be populated with a resource
+state. This is supported in two ways that can both be used:
+
+* By adding an entire tree of resources.
+* By invoking add/remove methods for each resource.
```
markuszoeller: Awesome, thanks. I think that's handy.

design/fake-client.rst
```diff
@@ -63,16 +67,22 @@ Design
...
-will use instead of zhmcclient.Session. The fake session object contains a
+will use instead of `zhmcclient.Session`. The fake session object contains a
 faked HMC that can represent all resources relevant for the zhmcclient.
-There fake session object implements the same API as zhmcclient.Session, so
-that the rest of the zhmcclient classes does not need to be changed.
-The get(), post() and delete() methods of the fake session class will
+
+This fake session object implements the same API as `zhmcclient.Session`, so
+that the rest of the zhmcclient classes do not need to be changed.
+The `get()`, `post()` and `delete()` methods of the fake session class will
```
markuszoeller: The (slight?) disadvantage of that approach is, that the testing surface is larger. We should give it a try though.

design/fake-client.rst
```diff
@@ -63,16 +67,22 @@ Design
...
 be redirected to operate on the faked HMC.
 
-The fake session object will have the ability to be "loaded" with an
-initial state of resources. The resource definitions can be provided as
-a dictionary of nested dictionaries, following this example::
+The fake session object will have the ability to be populated with a resource
+state. This is supported in two ways that can both be used:
```
markuszoeller: Supporting two different ways sounds like a lot more effort, but I think I need to see how this plays out in code.

PR #151, opened 2017-01-12
Fixes #152: Fixed assertion error in TimeStats testing on Windows.

### LINKED ISSUE #152: Appveyor fails in test_timestats
### Actual behavior

test_timestats fails only on Windows with:

```
self = <tests.unit.test_timestats.TimeStatsTests testMethod=test_measure_enabled>
    def test_measure_enabled(self):
        """Test measuring time with enabled keeper."""
    
        keeper = TimeStatsKeeper()
        keeper.enable()
    
        duration = 0.2
        delta = duration / 100
    
        stats = keeper.get_stats('foo')
        stats.begin()
        time.sleep(duration)
        stats.end()
    
        for _, stats in keeper.snapshot():
            self.assertEqual(stats.count, 1)
>           self.assertTrue(time_equal(stats.avg_time, duration, delta))
E           AssertionError: False is not true
tests\unit\test_timestats.py:133: AssertionError
```

### Expected behavior

passes

### Execution environment

* zhmcclient version: 0.9.0
* Operating system (type+version): Windows 7 in Appveyor CI

andy-maier:
Please review and merge.

Details:
* In test_timestats, improved the approach for checking time deltas such
  that the assertion error now shows the actual and expected delta, instead
  of just the mismatch between true and false.
* In test_timestats, increased the allowable time delta and the measured times
  in order to accomodate the fact that `time.time()` which is used by the `TimeStats`
  class, has only a 1/60 sec precision on Windows. This is a circumvention - the real
  solution would be to use some other service that returns high precision  timings
  on all supported operating systems.

### DIFF
--- a/tests/unit/test_timestats.py
+++ b/tests/unit/test_timestats.py
@@ -34,11 +34,11 @@ PRINT_HEADER_DISABLED = \
     "Disabled."
 
 
-def time_equal(t1, t2, delta):
+def time_abs_delta(t1, t2):
     """
-    Return True if two float values are nearly equal (as defined by delta).
+    Return the positive difference between two float values.
     """
-    return abs(t1 - t2) < delta
+    return abs(t1 - t2)
 
 
 class TimeStatsTests(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 14a04b10d945c8271f47d22e56d3e3c996d41672

---

leopoldjuergen:
+1

PR #159, opened 2017-01-16
Mocking support for zhmcclient, stage 1: Faked HMC.

andy-maier:
This is stage 1 of the mocking support according to the design described in PR #144. This stage provides a faked HMC. This stage is now complete and covers all resource types supported by the zhmcclient.

The stages have been split for easier reviewing. However, the two PRs should be merged at the same time once both have been approved.

A separate stage 2 in PR #160 will provide a faked Session object that links the zhmcclient package with the faked HMC.

Details:

- Added support for a faked HMC that provides a local in-memory representation of an HMC. Its API is intended to be used by a future faked Session object (which is the reason it is not compatible with the external zhmcclient API).


leopoldjuergen:
+1

PR #160, opened 2017-01-16
Mocking support for zhmcclient, stage 2: Faked Session (partially).

andy-maier:
This is stage 2 of the mocking support according to the design described in PR #144. 

Stage 1 is provided in PR #159.

The stages have been split for easier reviewing. However, the stages should be merged together once both have been approved.

At this point, the implementation of stage 2 of the mock support does not yet cover all operations supported by the zhmcclient. Missing are:

* For DPM mode:
```
    POST     /api/partitions/([^/]+)/operations/scsi-dump
    POST     /api/partitions/([^/]+)/operations/psw-restart
    POST     /api/partitions/([^/]+)/operations/mount-iso-image
    POST     /api/partitions/([^/]+)/operations/unmount-iso-image
    POST     /api/partitions/([^/]+)/hbas/([^/]+)/operations/reassign-
                  storage-adapter-port
    POST     /api/virtual-switches/([^/]+)/operations/get-connected-vnics
```
* For classic mode:
```
    POST     /api/cpcs/([^/]+)/operations/import-profiles
    POST     /api/cpcs/([^/]+)/operations/export-profiles
```

### DIFF
--- a/zhmcclient_mock/_hmc.py
+++ b/zhmcclient_mock/_hmc.py
@@ -1,0 +1,1014 @@
+# Copyright 2016 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+The `zhmcclient_mock` package provides a faked HMC with all resources that are
+relevant for the `zhmcclient` package. The faked HMC is implemented as a
+local Python object and maintains its resource state across operations.
+"""
+
+from __future__ import absolute_import
+
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
+import six
+# from six.moves.urllib.parse import urlparse, parse_qsl
+
+# TODO: Move the resources into their own files.
+
+__all__ = ['FakedBaseResource', 'FakedBaseManager', 'FakedHmc',
+           'FakedActivationProfileManager', 'FakedActivationProfile',
+           'FakedAdapterManager', 'FakedAdapter',
+           'FakedCpcManager', 'FakedCpc',
+           'FakedHbaManager', 'FakedHba',
+           'FakedLparManager', 'FakedLpar',
+           'FakedNicManager', 'FakedNic',
+           'FakedPartitionManager', 'FakedPartition',
+           'FakedPortManager', 'FakedPort',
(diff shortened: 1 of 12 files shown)


zhmcclient_mock/_hmc.py
```diff
@@ -0,0 +1,1006 @@
...
+        Parameters:
+
+          oid (string):
+            The object ID of the resource (e.g. value of the 'object-uri'
+            property).
+        """
+        uri = self._resources[oid].uri
+        del self._resources[oid]
+        del self._hmc._resources[uri]
+
```
preethipyp: find and findbyname are also frequently used methods which should be added

zhmcclient_mock/_hmc.py
```diff
@@ -0,0 +1,1006 @@
...
+        Parameters:
+
+          properties (dict):
+            Resource properties to be updated. Any other properties remain
+            unchanged.
+        """
+        self.properties.update(properties)
+
+
+class FakedBaseManager(object):
```
preethipyp: functions like create/delete which should internally add/remove the resources 
for example: _cpc.partition.create(properties) should be implemented in the FakeBasedManager to add a new Partition with properties to the model

zhmcclient_mock/_hmc.py
```diff
@@ -0,0 +1,1006 @@
...
+    @property
+    def ports(self):
+        """
+        The Port resources of this Adapter
+        (:class:`~zhmcclient_mock.FakedPort`).
+        """
+        return self._ports
+
+
+class FakedCpcManager(FakedBaseManager):
```
preethipyp: Some of the operations like start, stop are missing on this class.

preethipyp:
Overall the zhmcclient_mock looks great. But I see some basic functions like create, find, findbyname are missing. Is it expected to have this function mocked on the test case or the plan is to extend this framework to implement those functions too?

zhmcclient_mock/_hmc.py
andy-maier: Preethi, `find()` and `findall()` are supported.
The mock layer works such that its `FakedSession` is instantiated by the testcase using the mock layer instead of the real `Session`. For example, the find() function on CPCs is still available in the same way it would be with a real `Session`, but when going through the `FakedSession`, it becomes a `list()` on the `FakedCpcManager` of the `FakedHmc`. The bottom line is, the methods of the `FakedHmc` class are low level add/remove/list methods that are used internally by the mock layer, and therefore it is not meant to be used directly and therefore not compatible with the external API.

zhmcclient_mock/_hmc.py
andy-maier: Creation and deletion of e.g. partitions is supported. Again, the testcase uses the external zhmcclient API, just with a FakedSession instead of the real Session.

zhmcclient_mock/_hmc.py
andy-maier: Again, start() and stop() at the external APi will work. They become post() calls on the (internal) UriHandler. the handler for those is not yet implemented at this point, but when it is, it will simply change the status of the partition object in the FakedHmc object via direct properties update.

andy-maier:
Preethi, those functions are there. See my mail to you.

leopoldjuergen:
+1

Issue #161, opened 2017-01-17
AttributeError: 'Partition' object has no attribute 'cpc'"


leopoldjuergen:
### Actual behavior
```
vagrant@vagrant-ubuntu-trusty-64:/vagrant$ zhmc hba create P000AABB TestPart2 --name PCHID105 --adapter 'FCP 0105 A13B-08' --port 0 --device-number 4711
Traceback (most recent call last):
  File "/usr/local/bin/zhmc", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/zhmccli/_cmd_hba.py", line 110, in hba_create
    cmd_ctx.execute_cmd(lambda: cmd_hba_create(cmd_ctx, cpc, partition,
  File "/usr/local/lib/python2.7/dist-packages/zhmccli/_helper.py", line 134, in execute_cmd
    cmd()
  File "/usr/local/lib/python2.7/dist-packages/zhmccli/_cmd_hba.py", line 111, in <lambda>
    options))
  File "/usr/local/lib/python2.7/dist-packages/zhmccli/_cmd_hba.py", line 208, in cmd_hba_create
    adapter = partition.cpc.adapters.find(name=adapter_name)
AttributeError: 'Partition' object has no attribute 'cpc'
```

### Expected behavior
Successful creation of an HBA

### Execution environment

* zhmcclient version: 0.9.0
[...]

PR #162, opened 2017-01-17
Fix access 'cpc' from 'partition'.

leopoldjuergen:
Fix for AttributeError: 'Partition' object
has no attribute 'cpc'
Resolves issue #161

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -45,6 +45,10 @@ Released: not yet
 
 **Bug fixes:**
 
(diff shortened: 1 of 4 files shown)


leopoldjuergen:
+1

andy-maier:
-1
By design, only manager objects know their scoping resource.
The resources know only their manager, but not their scoping resource.
This keeps the number of references small, and it is only a very small inconvenience for users to have to go through the 'manager' attribute of a resource.

andy-maier:
+1
(Sorry, I misread the change when I posted my earlier review - deleted that one).

PR #166, opened 2017-01-19
Improved example lpar_operations

andy-maier:
Please review and merge.

Reason for this change was to have better displaying of the behavior that sometimes after LPAR activation, the status does not update for some time, or not at all.

Details:
 - Added retry loops when retrieving status, until desired status is achieved.
- Changed refresh from pull_full_properties() to using list(), because we just want to use the 'status' property.
- Removed "cpcstatus" input parameter, because when the CPC name is provided, it does not really make sense to filter by status in addition.

### DIFF
--- a/examples/lpar_operations.py
+++ b/examples/lpar_operations.py
@@ -96,8 +96,8 @@ print("Finding LPAR by name=%s ..." % lparname)
 # We use list() instead of find() because find(name=..) is optimized by using
 # the name-to-uri cache and therefore returns an Lpar object with only a
 # minimal set of properties, and particularly no 'status' property.
-# That would drive an extra "Get Partition Properties" operation when the
-# status property is accessed.
+# That would drive an extra "Get Logical Partition Properties" operation when
+# the status property is accessed.
 lpars = cpc.lpars.list(filter_args={'name': lparname})
 if len(lpars) != 1:
     print("Could not find LPAR %s in CPC %s" % (lparname, cpc.name))
--- a/examples/lpar_operations.py
+++ b/examples/lpar_operations.py
@@ -21,6 +21,7 @@ and activate/deactivate/load of an LPAR.
 import sys
 import logging
 import yaml
+import time
 import requests.packages.urllib3
 
 import zhmcclient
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

abrezovsky:
The delay on the updated status is certainly annoying, the solution on my scripts is to sleep for 5 seconds. Your solution is much more elegant.

With the performance issues of pull_full_properties, should we be suggesting using it to just get the status? Considering that something like the following is 150x faster?

`lpars = cec.lpars.list()
for aLpar in lpars:
    if aLpar.name == lparStr:
        status = aLpar.properties['status']
        break
if status == "":
    print("[Error] Lpar has disappeared...")
    sys.exit(1)`

andy-maier:
I agree that using `manager.list()` is preferrable. Will change it.

andy-maier:
I think we could go even further by having the loop on the desired status inside of the `lpar.activate()`, `lpar.deactivate()` and `lpar.load()` methods.

abrezovsky:
The question is, do we want to leave the verification that things went right to the user or handle it ourselves?

andy-maier:
The difficulty with doing it in the zhmcclient layer is async vs. sync. i think that by definition, we cannot do this for async mode, and the user needs to be aware of any deferrals w.r.t. status anyway. So if we cover for that in sync mode, I'm not sure we made it simpler. The sync and async behavior are now more apart than without covering for that.

So i tend to say we should not cover for that, but I'm not entirely sure about it.

examples/lpar_operations.py
```diff
@@ -82,66 +82,89 @@
...
-print("Accessing status of CPC %s ..." % cpc.name)
-status = cpc.get_property('status')
-print("Status of CPC %s: %s" % (cpc.name, status))
-
 print("Finding LPAR by name=%s ..." % lparname)
-try:
-    lpar = cpc.lpars.find(name=lparname)
-except zhmcclient.NotFound:
+# We use list() instead of find() because find(name=..) is optimized by using
+# the name-to-uri cache and therefore returns a Partition object with only a
```
leopoldjuergen: Why is the comment about Partitions if we are using LPArs here.

leopoldjuergen:
+1

Issue #148, opened 2017-01-11
Support LPAR OS message interactions


abrezovsky:
### Actual behavior

Interaction with LPAR OS messages is not currently supported.

### Expected behavior

Use Open OS Message Channel and/or Send OS Command operations to support interaction with LPAR OS messages.

HMC API Pages 513-517

### Execution environment

* zhmcclient version: 0.9.0
* Operating system (type+version): Ubuntu 16.10 (applies to any)

abrezovsky:
This would need some support for the Java Message Service. Would we need to find a package for JMS?

leopoldjuergen:
I think we have to implement the LPAR operation 'Open OS Message Channel' and combine it with example7.py. To connect to the API message broker we use STOMP in the example5.

abrezovsky:
Ah, OK, I didn't realize there was some flavor of that support in there already, I'll try to understand what's there already and then maybe start a new branch.

abrezovsky:
After looking at the jms_notifications example and how STOMP works, is the best solution  to just pass along the JMS topic to the client and allow them to set up STOMP themselves as in that example? A method in the Lpar class. I think that would be a good minimum.

Past that, we could implement certain aspects of the STOMP process within our wrapper and leave certain parts to the user?

I'm not exactly sure what that would be and/or how it would fit into our goal of providing a "thin wrapper".

andy-maier:
I have a suggestion on the design of the external API for this support:
1. The operation for opening the message channel should be a method on the `Partition` and `Lpar` classes, respectively, that returns an object of a new `MessageChannel` class.
2. The `MessageChannel` class has a method `get_next()` that waits for the next message on the JMS notification topic for the channel, and returns it to its caller once the next message has been received.

This way, the whole area of multithreading/multiprocessing/eventing/notification can be done in any way by the user, and the zhmcclient package does not need to select a particular multiprocessing or eventing model.

Comments, please.

abrezovsky:
I think what you are proposing would require the method to kick off a listener (see the example) in a new thread (or process). 

For now I'm going to implement an `open_os_message_channel()` method for `Partition` and `Lpar` to expose the HMC API at the very least.

andy-maier:
No, the open method would not have to start a thread.

abrezovsky:
I added in basic support through an open_os_message_channel method to the two classes. The pull request is open. I suppose we can keep this issue open if we feel we need to add more support?

[...]

PR #167, opened 2017-01-20
Added open_os_message_channel() method to lpar and partition classes

abrezovsky:
Added support for open-os-message-channel HMC API operation to the
lpar and partition classes. This is the minimum level of support for interaction with
operating system messages. The message channel will stay open until the API
session is terminated. #148 

Signed-off-by: Andrew Brezovsky <abrezovsky@gmail.com>

### DIFF
--- a/zhmcclient/_lpar.py
+++ b/zhmcclient/_lpar.py
@@ -298,3 +298,41 @@ class Lpar(BaseResource):
             self.uri + '/operations/load', body,
             wait_for_completion=wait_for_completion)
         return result
+
+    def open_os_message_channel(self, include_refresh_messages=True):
+        """
+        Open a JMS message channel to this LPAR's operating system, returning
+        the string "topic" representing the message channel.
+
+        Parameters:
+
+          include_refresh_messages (bool):
+            Boolean controlling whether refresh operating systems messages
+            should be sent, as follows:
+
+            * If `True`, refresh messages will be recieved when the user
+              connects to the topic. The default.
+
+            * If `False`, refresh messages will not be recieved when the user
+              connects to the topic.
+
+        Returns:
+
+          :term:`json object`:
+
+            Returns a JSON object with a member named ``topic-name``, a string
+            representing the os-message-notification JMS topic. The user can
+            connect to this topic to start the flow of operating system
+            messages.
+
+        Raises:
+
+          :exc:`~zhmcclient.HTTPError`
+          :exc:`~zhmcclient.ParseError`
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: bd36bb5521f7f90a27b3919c0b891862d49a97b9

---

andy-maier:
Some initial comments before the weekend:
1. The open..() method should not return the entire response body, but instead just the topic string. (This is just for consistency, we generally try to shield the user from request bodies, response bodies, and URIs).
2. This is really quite low level, asking the user to set up JMS communication themselves. With the `MessageChannel` class I described in issue #148, the user could get close support and would be shielded from the JMS protocol.

abrezovsky:
On 1, I had just been basing it off of how the other methods are handled, which all seem to return the response in its whole. In practice, it returns an OrderedDict with the 'topic-name' field containing the string, which isn't too bad. I can modify it to unwrap the string and return it directly if that's the consensus.

On 2, I saw this as a simple and clean way of adding an early method of support. I believe we need further discussion surrounding a `MessageChannel` class which we could implement in the future surrounding this method. I see nothing wrong with providing a low-level interface as a foundation.

andy-maier:
On 1: The only place I remember where the response is returned directly, is when methods that can be asynchronous (i.e. those that have a `wait_for_completion` argument such as `Cpc.start()`) return the job information in the asynchronous case. I think that was bad design on our behalf, and we should have created a `Job` object for that, just like we have resource objects for all resources. So my take is we should not use that "bad" design as a precedence. Bottom line, I'd really like to return just the notification topic string.

On 2: If you insist on returning the notification topic string in some way, we can still have a `MessageChannel` class as I proposed, that gets initialized with that notification topic string. That gives users the flexibility to use that class or to deal with the JMS transport protocol (STOMP or the other one)  themselves. Would that be an acceptable compromise to you?

abrezovsky:
This makes sense to me. I'll put another commit through that unwraps the string and then returns it directly. We can deal with the MessageChannel class in another branch.

andy-maier:
Great! I have a general purpose notification receiver ready for review, plus an example that uses your new method and the notification receiver.
I'll post that as a separate PR, and will adjust it to the unwrapped topic string return value once you change it.

andy-maier:
I created PR #168 on top of this PR, adding a JMS notification receiver, and an example displaying OS messages.

andy-maier:
+1 from my side.

leopoldjuergen:
+1

PR #182, opened 2017-02-02
Fixes #178, fixes #63: Redesigned return value of async. operation methods

### LINKED ISSUE #63: JSON objects returned by asynchronous methods are not convenient
The resource class methods invoking asynchronous HMC operations return the JSON object returned by the HMC if `wait_for_completion=False`. The only information returned in that object is the URI of the Job object.

Proposal: Change the return value to return a new `Job` object that is a resource object filled with just its URI, and that has a method to pull the job status.

### LINKED ISSUE #178: Doc says that partition.start() and partition.stop() should return None, but it retuns something
### Actual behavior
partiton.start() returns a Dict like this:

OrderedDict([(u'status', u'complete'), (u'job-status-code', 200), (u'job-results', OrderedDict()), (u'job-reason-code', 0)])

### Expected behavior
Along the docs [1], I would ave expected that it returns None

"Returns:	If wait_for_completion is True, returns None."
And the docs also says:
"start(wait_for_completion=True)" is the default

### Execution environment

* zhmcclient version: 0.9.0
* Operating system (type+version): ubuntu 14 x86

[1] http://python-zhmcclient.readthedocs.io/en/stable/resources.html#partitions

andy-maier:
Please review and merge.

**Note the incompatible change of the return value of async methods.**

Motivation:
- This fixes a discrepancy between the documentation that states `None` is returned in case of `wait_for_completion=True`, and the code that returned the response of the 'Query Job Status' operation as a JSON object. See issue #178.
- Because the return value changes anyway, this change also improves the return value in case of `wait_for_completion=False`, by returning an object of a new `Job` class.

Details:
- Changed the return type of all methods on resource classes that invoke asynchronous operations (i.e. all methods that have a `wait_for_completion` parameter), as follows:
  - For `wait_for_completion=True`, the JSON object in the 'job-results' field is now returned, or `None` if not present (i.e. no result data). Previously, the complete response was returned as a JSON object.
  - For `wait_for_completion=False`, a new `Job` object is now returned that allows checking and waiting for completion directly on the `Job` object. Previously, the whole response of the 'Query Job Status' operation was returned as a JSON object, and the job completion was checked on the `Session` object, and one could not wait for completion.
- Updated documentation, unit test cases for the `_session` module, and the examples `async_job_polling.py` and `jms_notifications.py` to accomodate the changed return values.
- Changed default of the `wait_for_completion` parameter on the `Session.post()` methods from `True` to `False`, in order to avoid superfluos timestats entries for the async method + completion polling, if the method was synchronous or asynchronous with `wait_for_completion=False`.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -27,11 +27,35 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Changed the return value of all methods on resource classes that invoke
+  asynchronous operations (i.e. all methods that have a `wait_for_completion`
+  parameter), as follows:
+
+  - For `wait_for_completion=True`, the JSON object in the 'job-results' field
+    is now returned, or `None` if not present (i.e. no result data).
+    Previously, the complete response was returned as a JSON object.
+
+  - For `wait_for_completion=False`, a new `Job` object is now returned that
+    allows checking and waiting for completion directly on the `Job` object.
+    Previously, the whole response of the 'Query Job Status' operation was
+    returned as a JSON object, and the job completion was checked on the
+    `Session` object, and one could not wait for completion.
+
+* Changed the default value of the `wait_for_completion` parameter of the
+  `Session.post()` method from `True` to `False`, in order to avoid
+  superfluos timestats entries. This method is not normally used by
+  users of the zhmcclient package.
+
 **Deprecations:**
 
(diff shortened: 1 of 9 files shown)


leopoldjuergen:
+1

PR #194, opened 2017-02-06
Added send_os_command method to partition and lpar

abrezovsky:
send_os_message allows the client to send messages to the object's operating
system. The method takes in a string for the command and and optional boolean
to indicate a priority. The method returns nothing because the HMC API returns
a response with no content. A status code 409 is returned from the API if the
message interface is not available.

Signed-off-by: Andrew Brezovsky <abrezovsky@gmail.com>

### DIFF
--- a/zhmcclient/_lpar.py
+++ b/zhmcclient/_lpar.py
@@ -351,3 +351,37 @@ class Lpar(BaseResource):
         result = self.manager.session.post(
             self.uri + '/operations/open-os-message-channel', body)
         return result['topic-name']
+
+    def send_os_command(self, os_command_text, is_priority=False):
+        """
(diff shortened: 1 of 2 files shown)


andy-maier:
Andrew, I have only these minor comments:
- Please change `true` and `false` in the description of `is_priority` in the docstrings to `True` and `False` (still in backquotes).

Also, we changed the setup of the CI systems, which is the reason the "continuous-integration/appveyor/branch" check shows up as under way, but it was just added (again). This will automatically take care of itself once you push your update to the PR.

andy-maier:
+1

leopoldjuergen:
+1

leopoldjuergen:
Created a new issue #203

PR #195, opened 2017-02-06
Fixes #6, Fixes #191: Timeout support for async methods; deferred status polling for LPAR

### LINKED ISSUE #6: Timeout support
Add support for timing out HTTP requests.

### LINKED ISSUE #191: Deferred status transition with Lpar.activate() and deactivate()
When the `Lpar.activate()` and `deactivate()` methods complete (e.g. by returning after being called with `wait_for_completion=True`), the LPAR state may not have reached the desired state, and a subsequent status retrieval still reports the old state.

This does not happen always, and the exact circumstances are not fully understood at this point.

As a result, users of the zhmcclient have to code around this and loop on status retrieval until the desired state is reached.

The zhmcclient could take care of that by performing the stats polling inside of these methods for the `wait_for_completion=True` case, making life easier for the user. In the `wait_for_completion=False` case, adressing this is more complex, but not impossible, given that we have PR #182: The async method could return a new subclass of `Job` that implements its check and wait methods such that completion includes polling for the desired status.

Discussion points:
* Should we do this at all?
* Should a timeout be added? (see Issue #6)

andy-maier:
Please review and merge.

Please focus on the changes at the API, more than on the implementation code.

Good questions would be, for example:

- The new `TimeoutError` class is used for both async operation timeouts and status polling timeouts, and it does not have a machine-readable way to distinguish these cases. This affects only the `Lpar.activate/dectivate/load()` methods. Is this good enough?
- Are the default timeouts right (see the new `_constants.py` module)?
- The use of `Lpar.activate/dectivate/load()` with `wait_for_completion=False` returns a Job object that can handle the waiting for job completion, but not the waiting for deferred status update. On the one hand, this may be what you expect from an asynchronous job, on the other hand, it is not consistent with the `wait_for_completion=True` case, where both is done by the zhmcclient library.

Details, from the commit message:

- Added exception class `TimeoutError` that is used to indicate operation timeout and status polling timeout.
- Added support for deferred status polling to the `Lpar.activate/deactivate/load()` methods.i
  Background: The HMC operations issued by these methods exhibit show deferred status ibehavior, which means that it takes a few seconds after successful completion of the async job that executes the operation, until the new status can be observed in the 'status' property of the LPAR resource.
  These methods will poll the LPAR status until the desired status value is reached. A status timeout can be specified via a new `status_timeout` parameter to these methods, which defaults to 1h. The long timeout duration means this is just a last resort.
- Added operation timeout support to `Session.post()` and to all resource methods with a `wait_for_completion` parameter (i.e. the asynchronous methods). The operation timeout on the async methods can be specified via a new `operation_timeout` parameter, which defaults to no timeout.
- Added a new module `_constants` to define the default timeout values for the operation timeout and the status timeout.

### DIFF
--- a/design/network-errors.rst
+++ b/design/network-errors.rst
@@ -1,0 +1,380 @@
+.. Copyright 2017 IBM Corp. All Rights Reserved.
+..
+.. Licensed under the Apache License, Version 2.0 (the "License");
+.. you may not use this file except in compliance with the License.
+.. You may obtain a copy of the License at
+..
+..    http://www.apache.org/licenses/LICENSE-2.0
+..
+.. Unless required by applicable law or agreed to in writing, software
+.. distributed under the License is distributed on an "AS IS" BASIS,
+.. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+================================================
+Design for handling network errors in zhmcclient
+================================================
+
+The zhmcclient package uses the Python
+`requests <https://pypi.python.org/pypi/requests/>`_ package for any HMC REST
+API calls, and the Python `stomp <https://pypi.python.org/pypi/stomp/>`_
+package for handling HMC JMS notifications.
+
+This design document covers error handling using the `requests` package.
+
+Some facts
+==========
+
+Layering of components
+----------------------
+
+The following picture shows the layering of components w.r.t. networking:
+
+::
+
+  +---------------------+
+  |                     |
+  |  application (py)   |
+  |                     |
+  +---------------------+
+  |                     |
+  |   zhmcclient (py)   |
+  |                     |
+  +---------------------+
+  |                     |
+  |    requests (py)    |
+  |                     |
+  +---------------------+
+  |                     |  - Handles HTTP connection keep-alive and pooling
+  |    urllib3 (py)     |  - Handles HTTP basic/digest authentication
+  |  (part of requests) |  - Handles HTTP request retrying
+  +---------------------+
+  |                     |
+  |    httplib (py)     |
+  |                     |
+  +---------------+     |
+  |               |     |
+  |  socket (py)  |     |
+  |               |     |
+  +---------------+-----+
+  |    _socket (py)     |
+  |                     |
+  |  _socketmodule.so   |
+  |                     |
+  +---------------------+  - Handles read timeouts ??
+  |     socket API      |  - Handles connection timeouts ??
+  |    TCP/IP stack     |  - Handles SSL certificate verification ??
+  |                     |  - Handles TCP packet retransmission
+  +---------------------+
+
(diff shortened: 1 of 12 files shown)


abrezovsky:
+1

andy-maier:
+1

andy-maier:
Resetting this to [WIP] for the moment, until the design of PR #195 has been reviewed, in order to figure out how these two come together.

andy-maier:
This PR needs to be updated, by rebasing it upon PR #206, and by supporting the two timeouts via the RetryTimeoutConfig class at the session level.

andy-maier:
The PR has been rebased and extended as described in this PR and in iisue #191.

abrezovsky:
I would really love for this to be in the next release. As it solves a fundamental issue with fully fledged automation.

leopoldjuergen:
Get problem with read timeout with example get_partial_and_full_properties.py
I think we have to adjust the timeouts.
```
(venv) vagrant@vagrant-ubuntu-trusty-64:/vagrant/examples$ ./get_partial_and_full_properties.py hmccreds.yaml 

Example shows how to get partial and full properties for CPCs
and for LPARs of a CPC.

Using HMC XXX.XXX.XXX.XXX with userid ensadmin ...
Listing CPCs with full_properties=False ...
Duration: 0:00:00.586465
Number of properties of CPC P0ZHMP02: 3
Number of properties of CPC P0000P30: 3
Listing CPCs with full_properties=True ...
Traceback (most recent call last):
  File "./get_partial_and_full_properties.py", line 84, in <module>
    cpcs = cl.cpcs.list(full_properties)
  File "<decorator-gen-18>", line 2, in list
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_logging.py", line 132, in wrapper
    result = func(*args, **kwargs)
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_cpc.py", line 149, in list
    resource_obj.pull_full_properties()
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_resource.py", line 204, in pull_full_properties
    full_properties = self.manager.session.get(self._uri)
  File "<decorator-gen-4>", line 2, in get
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_logging.py", line 132, in wrapper
    result = func(*args, **kwargs)
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_session.py", line 601, in get
    _handle_request_exc(exc)
  File "/vagrant/venv/local/lib/python2.7/site-packages/zhmcclient/_session.py", line 63, in _handle_request_exc
    raise ConnectionError(_request_exc_message(exc), exc)
zhmcclient._exceptions.ConnectionError: HTTPSConnectionPool(host='XXX.XXX.XXXX.XXX', port=6794): Read timed out. (read timeout=30)
```

leopoldjuergen:
+1

PR #200, opened 2017-02-13
Fixes #199: Added minimum version 4.0.0 to 'decorate' dependency.

### LINKED ISSUE #199: Minimum version for decorator dependency is missing
### Actual behavior

The zhmcclient package uses the "decorator" Python package, and correctly specifies that package as a dependency in its `requirements.txt` file. However, the zhmcclient package uses the `decorator.decorate()` function from that package, which was introduced only in its version 4.0.0.

### Expected behavior

The zhmcclient package should specify `>=4.0.0` as a version requirement on the decorator dependency.

### Execution environment

* zhmcclient version: any
* Operating system (type+version): any

andy-maier:
Please review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -50,6 +50,9 @@ Released: not yet
 
 **Bug fixes:**
 
+* Added a minimum version requirement `>=4.0.0` for the dependency on the
+  "decorate" Python package (issue #199).
+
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #204, opened 2017-02-28
Added automatic devno setting for NIC, HBA, VF in mock support

andy-maier:
Details:
- Added an IdPool class that can allocate and free integer ID values. This is used for devno allocation in faked partitions.
- Added unit test cases for the new IdPool class.
- Added automatic setting of the 'device-number' property when not specified by the caller, for faked HBA, NIC and VF resources, in the value range of 0x8000 to 0xFFFF.


leopoldjuergen:
+1

PR #206, opened 2017-03-01
Improved handling of network errors

andy-maier:
This PR is now ready for code review and merging. It includes a design document in `design/network-errors.rst`.

It attempts to handle all network-related errors in the zhmcclient layer, if possible (e.g. timeouts, retries), so that the user of the package does not need to care about retries, and consider any raising of `zhmcclient.ConnectionError` to be not resolvable by user-level retrying.

The timeouts for async operation completion and LPAR status transition completion that were in an intermediate version of this PR have been removed again, so that this PR is now independent of PR #195 and needs to go in first. PR #195 will be changed to be based on this PR.


andy-maier:
+1

PR #219, opened 2017-03-21
Fixed display of CLI spinner in output

andy-maier:
The CLI spinner is correctly stopped but still appears with one character in the output. The reason for that is that it is still spinning while the output is written, and is stopped only after that. So the printing of the spinner character and the printing of the actual output interfere.

This change stops the spinner already before the output is written, so the spinner code itself cleans up the spinner characters.

Please review and merge.


andy-maier:
+1

leopoldjuergen:
+1

PR #229, opened 2017-03-31
Fixes #228: Added support for 'error' field in 'job-results'

### LINKED ISSUE #228: Starting a paused partition fails
### Actual behavior
This is the expected behavior by DPM (confirmed that) - but it's really strange. For nova-dpm we need ot fix it - the question is, if we fix it in zhmcclient or in nova-dpm.

- do an inband shutdown of a partition (e.g. via linux shell:  shutdown -P now)
- partition state switches "Paused"
- try to start that partition using "Daily - Start"

This results in the following error on the UI
"STS00005E
One or more targeted partitions are already started. Only partitions in the Stopped state can be started.
To restart partitions, first run the Stop task, and then try the operation again."

The cli does not care about this issue at all, it just leaves the partition state to paused:
$ zhmc partition list P000S67B
/+-------------------------------------------------------------+------------+
| name                                                        | status     |
|-------------------------------------------------------------+------------|
| andreas_cloud_init_tests                                    | paused     |

$ zhmc partition start P000S67B andreas_cloud_init_tests
|Partition andreas_cloud_init_tests has been started.

$ zhmc partition list P000S67B
/+-------------------------------------------------------------+------------+
| name                                                        | status     |
|-------------------------------------------------------------+------------|
| andreas_cloud_init_tests                                    | paused     |

The python binding seems to behave like the cli. It returns immediately even if wait_for_completion ==True. There's no indication that the operation failed.

### Execution environment

* zhmcclient version: 0.9.0

andy-maier:
This improves the `HTTPError` that is raised for a job that completed in error, by also supporting the `error` field in the `job-results` object of the response. So far, only the `message` field was supported. The HMC API book is silent as to what the fields are on the `job-results` object when a job completes in error.

For more details, see the commit message.

Please review and merge.

### DIFF
--- a/zhmcclient/_session.py
+++ b/zhmcclient/_session.py
@@ -824,7 +824,7 @@ class Session(object):
             elif result.status_code == 202:
                 result_object = _result_object(result)
                 job_uri = result_object['job-uri']
-                job = Job(self, job_uri)
+                job = Job(self, job_uri, 'POST', uri)
                 if wait_for_completion:
                     return job.wait_for_completion(operation_timeout)
                 else:
@@ -954,7 +954,7 @@ class Job(object):
     completion.
     """
 
-    def __init__(self, session, uri):
+    def __init__(self, session, uri, op_method, op_uri):
         """
         Parameters:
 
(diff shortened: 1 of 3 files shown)


zhmcclient/_session.py
```diff
@@ -935,7 +935,7 @@ class Job(object):
     completion.
     """
 
-    def __init__(self, session, uri):
+    def __init__(self, session, uri, oper_method, oper_uri):
```
leopoldjuergen: I would like to see 'op' instead of 'oper' as an abbreviation for operation.
Why do you not use 'op_http_method' and 'op_uri'

andy-maier: Agreed. Has been changed in latest commit.

andy-maier:
Rebased.

leopoldjuergen:
+1

PR #230, opened 2017-03-31
Fixes #225: CLI: Added -p/--password option

### LINKED ISSUE #225: Pass password to zhmc on the command line
Add a switch –p to pass the password on the command line.

Example:
zhmc -o plain -h ${HMC} -u ${HMCUSER} -p ${HMCPASSWD} lpar list ${CECDCP}

The -p option should be considered the least secure of all password options.

andy-maier:
Please review and merge.

Note that we now have both ZHMC_PASSWORD and ZHMC_SESSION_ID as environment variables, and either can be used to avoid the password prompt and use of the new password option. Maybe we need to clean that up a bit.

Details:
- Added a `-p` / `--password` option to specify the HMC password.
- Changed interface to password retrieval function invoked by 'Session'; it now has host and userid as arguments. This needed to be done because host and userid were passed by closure, while now they are passed explicitly. This also helps when users start using this feature.
- Added a `__repr__()` function for `CmdContext`, for better debugging.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -27,6 +27,10 @@ Released: not yet
 
 **Incompatible changes:**
 
+* The password retrieval function that can optionally be passed to
+  ``Session()`` has changed its interface; it is now being called with host and
+  userid. Related to issue #225.
+
 **Deprecations:**
 
 **Bug fixes:**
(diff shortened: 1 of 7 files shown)


leopoldjuergen:
+1

PR #238, opened 2017-04-12
Fixed package dependencies for minimum pkg level

andy-maier:
Please review and merge if the test runs succeed.

Background:

  There were many indirect package dependencies that were not
  listed in the requirements and version constraint files.
  This resulted in an install failure when testing with
  PACKAGE_LEVEL=minimum, because the specified old versions of
  the packages that were listed, were mixed with the latest
  versions of the packages that were not listed.

Details of changes:

- Added all missing package dependencies as of 2016-12-08, with
  their versions as of that date as a minimum.
- Changed the minimum versions of already existing package
  dependencies to the versions as of 2016-12-08. This resulted
  sometimes in increasing the minimum version, and sometimes in
  decreasing it.

Signed-off-by: Andreas Maier <maiera@de.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,12 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "pypy"
+#  - "pypy"        # PyPy 2.5.0, Python 2.7.8
+  - "pypy-5.3.1"  # PyPy 5.3.1, Python 2.7.10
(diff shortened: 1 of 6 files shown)


andy-maier:
+1 - it finally works in all environments.

leopoldjuergen:
+1

PR #245, opened 2017-04-15
Fixes #210: Wrote Concepts chapter

### LINKED ISSUE #210: Verify multi-threading support
The zhmcclient so far was not designed with multi-threading capability in mind. It is not clear whether there are any issues. The OpenStack Nova DPM driver uses the same Session object in multiple threads.

andy-maier:
This chapter has a section "multi-threading considerations" in which it states that each session, client and notification receiver object can be used by only one thread at a time. It therefore addresses issue #210.

Please review and merge.


andy-maier:
+1

leopoldjuergen:
+1

Issue #249, opened 2017-04-20
Read retries cause multiple DELETE requests


andy-maier:
### Actual behavior

The default values for the retry / timeout configuration specify `read_timeout=300` (sec) and `read_retries=3`.

This configuration causes the "requests" module to resend DELETE requests if the processing takes longer than the `read_timeout`.

This happens for example for the `Partition.delete()` method for a partition with many NICs that had been started and stopped.

### Expected behavior

A DELETE request should not automatically be re-requested after it is already being processed.

A short term mitigation is to change the default configuration to `read_retries=0`, `read_timeout={some large value}`.

A better solution needs to be discussed.

### Execution environment

* zhmcclient version: 0.12.0.dev3
* Operating system (type+version): RHEL 7.2
* HMC version: 2.13.1
* CPC version: 2.13.1

andy-maier:
Mitigated by PR #250.

andy-maier:
I did some more research on the question what the final solution should be:

According to the HTTP 1.1 spec, DELETE is an idempotent method, and idempotency is defined such that a second invocation of the same method does not change the resource state of the system. It may very well behave differently in its HTTP response.

For example, a first DELETE on a resource will change the resource state such that the resource is no longer there, and returns success in its response. A second DELETE on the same resource also leaves the system in a state where that resource does not exist, but returns "not found".

This is pretty much what the HMC implemented for Delete Partition: The second DELETE gets a "busy" error back, because the first DELETE is still in progress. It probably would return "not found" if the retry was performed just when the processing of the first DELETE was complete.

So the HMC behaves according to the HTTP spec for this case.

The `requests` module also behaves according to the HTTP spec when it resends the DELETE request, because DELETE is an idempotent HTTP method.

Nevertheless we do not want to return an error back to the zhmcclient users in such a case.

So the mitigation of turning off read retries may very well end up to be the final solution.

One option still to be investigated is the use of the `method_whitelist` parameter when creating the urllib3 `Retry` object.

andy-maier:
The `method_whitelist` argument of `urllib3.Retry()` allows restricting the resending of a request upon timeout during reading of a response to certain HTTP methods, i.e. we could restrict it to just GET.

However, I did not see a single case yet where the HMC did not respond to a GET (or any other method) at some point. So the only reason to have a read timeout and read retry configured would be to have an additional safety net that guarantees that the zhmcclient layer returns to its caller at some point.

[...]

PR #250, opened 2017-04-20
Mitigated issue with read retries and DELETE

andy-maier:
This PR mitigates issue #249. A final fix may or may not come, so this PR should be integrated in any case.

Please review and merge.

Details:
- This change sets the default values for the retry / timeout  configuration to read_timeout=3600 and read_retries=0. This  was needed in order to help with the issue that the 'requests' module re-requests a DELETE request another time if the first request is being processed but times out on the read timeout because it takes longer. The Delete Partition operation is such a candidate, at least when the partition got many NICs added and was started and stopped.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -38,6 +38,12 @@ Released: not yet
 * Fixed that the defaults for memory for the ``zhmc partition create`` command
   were ignored (issue #246).
 
+* The default values for the retra / timeout configuration for a session has
+  been changed to disable read retries and to set the read timeout to 1 hour.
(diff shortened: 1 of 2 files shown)


andy-maier:
I tested that with the new settings, a partition with 200 NICs can be deleted without a problem. That was the scenario that lead to the retry issue described in issue #249 when using the previous settings.

Therefore, +1.

andy-maier:
Rebased - waiting for CI runs to complete.

leopoldjuergen:
+1

PR #252, opened 2017-04-26
Fixes #249: Restricted read retries to just HTTP GET

### LINKED ISSUE #249: Read retries cause multiple DELETE requests

andy-maier:
This PR is proposed as the final solution to issue #249. See the discussion in that issue.

Please review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -38,11 +38,10 @@ Released: not yet
 * Fixed that the defaults for memory for the ``zhmc partition create`` command
   were ignored (issue #246).
 
-* The default values for the retra / timeout configuration for a session has
+* The default values for the retry / timeout configuration for a session has
   been changed to disable read retries and to set the read timeout to 1 hour.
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

PR #255, opened 2017-04-28
Fixes #253: Fixed stale name-to-URI cache and other improvements

### LINKED ISSUE #253: Stale partition URL after delete/create partition
### Actual behavior

If a "Create NIC" operation is issued on a partition that was just deleted and created again, it fails with HTTP status 404, reason 1, message "Not found or not authorized".

The sequence of HMC operations for reproducing the failure is:

- Logon
- List CPCs (and pick one)
- List Partitions (filter by name)
- Delete Partition
- Create Partition
- List partitions (filter by name)
- Create NIC on the resulting partition -> fails with 404,1

The WS API log on the HMC shows that the Create NIC operation in its target URL uses the partition ID of the partition resource that was deleted, when it should have used the newly created partition resource.

### Expected behavior

Use the partition ID of newly created partition, instead.

### Execution environment

* zhmcclient version: 0.12.0
* HMC version: 2.13.1
* CPC version: 2.13.1

andy-maier:
Please review and merge.

Details:
* Fixed that resource creation, deletion, and resource property updating now properly updates the resource name-to-URI cache in the zhmcclient that is maintained in the `*Manager` objects. As part of that, the `BaseManager` init function got an additional required argument `session`, but because creation of manager objects is not part of the external API, this should not affect users. See issue #253.
* The `update_properties()` method of all Python resource objects now also updates the properties of that Python resource object with the properties provided by the user (in addition to issuing the corresponding Update Properties HMC operation. This was done because that is likely the expectation of users, and we already store user-provided properties in Python resource objects when creating resources so it is now consistent with that. This came up as part of issue #253.
* As part of fixing the name-to-URI cache, a new attribute `name_uri_cache_timetolive` was added to class `RetryTimeoutConfig`, which allows controlling after what time the name-to-URI cache is automatically invalidated. The default for that is set in a new `DEFAULT_NAME_URI_CACHE_TIMETOLIVE` constant. Also, the `*Manager` classes now have a new method `invalidate_name_uri_cache()` which can be used to manually invalidate the name-to-URI cache, for cases where multiple parties (besides the current zhmcclient instance) change resources on the HMC. This came up as part of issue #253.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -43,9 +43,34 @@ Released: not yet
   In addition, read retries are now restricted to HTTP GET methods, in case
   the user enabled read retries. See issue #249.
 
+* Fixed that resource creation, deletion, and resource property updating now
+  properly updates the resource name-to-URI cache in the zhmcclient that is
+  maintained in the `*Manager` objects. As part of that, the `BaseManager`
+  init function got an additional required argument `session`, but because
+  creation of manager objects is not part of the external API, this should not
+  affect users. See issue #253.
+
 **Enhancements:**
 
 * Added content to the "Concepts" chapter in the documentation.
+
+* The `update_properties()` method of all Python resource objects now also
+  updates the properties of that Python resource object with the properties
+  provided by the user (in addition to issuing the corresponding Update
+  Properties HMC operation. This was done because that is likely the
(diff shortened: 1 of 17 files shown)


andy-maier:
+1

leopoldjuergen:
+1

PR #257, opened 2017-04-28
Fixes #256: Fixes in unit tests for update_properties() and delete()

### LINKED ISSUE #256: Unit tests for update_properties() and delete() assume incorrect return values
The unit tests for the `update_properties()` and `delete()` methods of most Python resource classes assume incorrect return values from these functions (from before one of the changes in this area). This applies to a number of these method calls, and one needs to investigate which ones exactly.

andy-maier:
Please review and merge.

Details:
- Fixed incorrect assumptions about their method return values. See issue #256.
- Fixed incorrectly returned response bodies for mocked DELETE and POST (for update), and replaced that with status 204 (no content). This came up as part of fixing issue #256.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -53,6 +53,15 @@ Released: not yet
   creation of manager objects is not part of the external API, this should not
   affect users. See issue #253.
 
+* In the unit testcases for the `update_properties()` and `delete()` methods of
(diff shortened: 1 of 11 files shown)


andy-maier:
+1

PR #266, opened 2017-05-04
Fixes #264: Fixed find(name) on element resources.

### LINKED ISSUE #264: find(name) on element-kind of resource classes raises NotFound for existing resource
### Actual behavior

`nics.find(name)` raises `NotFound` for an existing NIC resource.

This is caused by the fact that the `list()` method for NICs and other element-kind of resources returns Python resource objects that do not have a name property, so the operation that is supposed to populate the name-URI-cache does not add anything. Because of the optimization in find(name) to always use the name-URI-cache, it relies on that cache being populated.

This issue applies to all element-kind of resource types (NIC, HBA, VF, Port).

### Expected behavior

Resource should be found.

### Execution environment

* zhmcclient version: 0.12.1.dev19
* Operating system (type+version): RHEL 7.3
* HMC version: 2.13.1
* CPC version: 2.13.1

andy-maier:
Please review and merge.

From the commit message:

The find(name) method on element-kind of resources raised NotFound for existing resources. This was caused by the fact that the list() method for element-kind of resources returns Python resource
objects that do not have a "name" property, so the operation that is supposed to populate the name-URI-cache does not add anything. Because of the optimization in find(name) to always use the
name-URI-cache, it relies on that cache being populated. See issue #264.

This change fixes that by introducing knowledge into the manager class about whether the list() method for a resource class populates the "name" property. If that is not the case, the refresh() method of the name-to-URI cache sets the 'full_properties' flag when calling the list() method. For the element-kind of resource classes, this drives Get {Resource} Properties HMC operations to populate the cache, but afterwards, the cache is utilized. This has been done for all element-kind of resource classes, i.e. NICs, HBAs, VFs, Ports.

In addition, the unit tests for element-kind of resource classes have been extended with test cases for filtering with name and with element-id.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -62,6 +62,9 @@ Released: not yet
   DELETE and POST (for update), and replaced that with status 204 (no content).
   This came up as part of fixing issue #256.
 
+* Fixed that ``find(name)`` raised ``NotFound`` for existing resources, for
+  resource types that are elements (i.e. NICs, HBAs, VFs, Ports) (issue #264).
+
 **Enhancements:**
 
 * Added content to the "Concepts" chapter in the documentation.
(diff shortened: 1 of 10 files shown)


andy-maier:
+1

PR #282, opened 2017-05-19
Remove empty list items in change log

andy-maier:
This change removes all empty list items in the change log for all releases. It also updates the description on how to make a release, by adding an action item toremove the empty list items in the change log section of the current release.


leopoldjuergen:
+1

PR #300, opened 2017-05-26
CLI: Improvements for dealing with LPAR status 'exceptions'.

andy-maier:
As long as this PR is not yet merged, it can be installed via:

    pip install git+https://github.com/zhmcclient/python-zhmcclient.git@andy/fix-status-timeout

This change provides some improvement for the zhmc CLI when dealing with the LPAR status "exceptions" when activating, deactivating or loading the LPAR.

The main change is to add a command line option `--allow-status-exceptions` for the `lpar activate/deactivate/load` commands. Setting this option causes the LPAR status "exceptions" to be considered an additional valid end status.

Some other changes were needed as part of that:
* Allowed `None` as a value for the `load_parameter` argument of `Lpar.load(), and changed the default to be `None` (which does not change the behavior).
* Added actual status, desired statuses and status timeout as attributes to the `StatusTimeout` exception, for programmatic processing by callers. This is not yet exploited in the zhmc CLI, but it can already be exploited by users of the zhmcclient API.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -56,6 +56,18 @@ Released: not yet
   waiting for reaching the desired LPAR status after the HMC operation
   'Activate LPAR' or 'Deactivate LPAR' has completed.
 
+* Allow ``None`` as a value for the ``load_parameter`` argument of
+  ``Lpar.load()``, and changed the default to be ``None`` (the latter change
+  does not change the behavior).
+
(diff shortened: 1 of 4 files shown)


andy-maier:
Based upon feedback of the user who had the issue with LPAR status "exceptions", I have merged this PR.

PR #303, opened 2017-05-29
Added support for selecting the format of CLI error messages

andy-maier:
Details:
- Added `str_def()` methods to all zhmcclient exceptions. These methods return a string that is suited for being processed by scripts that invoke the zhmc CLI.
- Added a command line option `-e` / `-error-format` to the zhmc cli, which controls the format of any error messages. The `-e def` option selects the format returned by the `str_def()` methods.


leopoldjuergen:
+1

PR #307, opened 2017-05-31
Added wait_for_status() methods to Lpar and Partition.

andy-maier:
Details:
- The `wait_for_status()` method for the `Lpar` class was already available
  as an internal method, which was simply externalized.
- For the `Partition` class, a similar method has been added.
- The reason for these methods is that they make it much easier to ensure
  a particular operational status.


leopoldjuergen:
+1

leopoldjuergen:
rebased

PR #324, opened 2017-06-26
Moved PyLint invocation into its own 'pylint' make target.

andy-maier:
Previously, it was part of the 'check' make target. The motivation for this change is to shorten the CI runs.


andy-maier:
Rebased.

andy-maier:
+1

leopoldjuergen:
+1

PR #325, opened 2017-06-26
Fixes #57: Corrected parameters of Partition.mount_iso_image()

### LINKED ISSUE #57: Incorrect parameters for Partition.mount_iso_image()
There is a number of issues in the interface and implementation of the `Partition.mount_iso_image()` method:
- The method is currently defined to have input properties. However, the HMC operation 'Mount ISO Image' has no input properties that would be put into the request body, but instead the request body is the binary content of the image file.
- For this HMC operation, the `Content-Type` header must be set to `application/octet-stream` because the request body is binary data, but the current implementation does not set that header.
- This HMC operation has two query parameters (in the URL), which are not supported by the current interface and implementation of the method.

andy-maier:
Now ready for review and merge. It works now, if one satisfies the requirements for the image_name property which are now properly documented.

One more thing: Because this operation takes a long time, it would be nice  to be able to register a call back function for monitoring the upload progress. That is not yet in the current code, and could be added compatibly with an optional parameter.

Note, this is an incompatible change but that was unavoidable.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -30,6 +30,9 @@ Released: not yet
   key/value tuples to a dictionary. This is more flexible and reduces the
   number of data structure conversions in different scenarios. See issue #269.
 
+* Changed the arguments of ``Partition.mount_iso_image()`` incompatibly,
+  in order to fix issue #57.
+
 **Deprecations:**
 
 **Bug fixes:**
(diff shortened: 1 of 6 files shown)


andy-maier:
Rebased.

andy-maier:
+1

andy-maier:
Rebased.

PR #330, opened 2017-06-29
Added support for Get Inventory operation

andy-maier:

**Does not yet contain unit test code!**

Use the `examples/get_inventory.py` script to try it out.


andy-maier:
+1

PR #332, opened 2017-07-01
Added properties for max partitions to Cpc and for max crypto domains to Adapter

andy-maier:

The max domains property in the Adapter class has unit test code.
The max partitions property in the Cpc class does not have unit test code because it has not yet been converted to use the zhmcclient mock support. It will be added once that happens.


andy-maier:
+1

andy-maier:
Rebased.

leopoldjuergen:
+1

PR #336, opened 2017-07-03
Created tests/unit/utils.py with shared code between the testcases

andy-maier:
This PR will be used by multiple other PRs on testcases.
Please review and merge.

At this point, the module has only one function `assert_resources()` which checks a list of resources against an expected list of (real or faked) resources.


leopoldjuergen:
+1

PR #338, opened 2017-07-03
vlan-id must be an integer

leopoldjuergen:
Today,  the vlan-id is passed as a string to
the HMC WS API by the CLI. The HMC returns a
HTTPError: 400,7 because it expects that the
vlan-id is an integer. This change reads the
vlan-id as a integer instead as a string.


andy-maier:
+1

PR #339, opened 2017-07-03
Mock: Fixed FakedBaseResource init when properties are empty

andy-maier:
Ready for review and merge.


andy-maier:
This solves the problem that the ID and URI were not defaulted when properties was empty or None.

PR #345, opened 2017-07-04
Added update_properties() to Cpc and Lpar.

andy-maier:
Please review and merge.
The testcase migration changes will depend on this change.

**Note:** Unit test code will be added when migrating these resource types to py.test and zhmcclient mock support.


andy-maier:
+1

PR #346, opened 2017-07-04
Fixed updating the name-to-URI cache when renaming.

andy-maier:
Please review and merge.
The testcase migration changes will depend on this change.


andy-maier:
+1

PR #357, opened 2017-07-05
Docs: Improved Nic/Hba.create() descriptions.

andy-maier:
Ready to be merged. Doc-only!

Details:
- Improved the description of the Nic and Hba create() method to explain how the backing adapter port is specified, and how it can be determined from adapter name and port index. This is specifically interesting for OSa and HiperSockets adapters, because the virtual switch that is needed, is not familiar to users of the HMC GUI.


andy-maier:
+1

leopoldjuergen:
+1

PR #373, opened 2017-07-14
Juergen/metrics - again

andy-maier:


andy-maier:
WIll be done by PR #374.

PR #383, opened 2017-07-20
CLI support for processing weights

leopoldjuergen:
Follwing processing weights can be modified
during 'partition create' and 'partition
update':

- initial-cp-processing-weight
- initial-ifl-processing-weight
- maximum-cp-processing-weight
- maximum-ifl-processing-weight
- minimum-cp-processing-weight
- minimum-ifl-processing-weight


leopoldjuergen:
Added an entry in the change log for these new properties, and for the new properties in the previous PR.

PR #391, opened 2017-08-07
Add config file for Code Climate tool

leopoldjuergen:
Code Climate is a fully-configurable static analysis
tool. This change will provide a .codeclimate.yml
configuration file which enables the two Code Climate
engines Radon (https://github.com/rubik/radon) and Duplication
(https://github.com/codeclimate/codeclimate-duplication).
Radon is a Python tool that computes various metrics like
cyclomatic complexity from the source code.
Duplication analyzes code for structural similarities.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,15 @@
+---
+engines:
+  duplication:
+    enabled: true
+    config:
+      languages:
+        - python
+  radon:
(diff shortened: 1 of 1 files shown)


leopoldjuergen:
I have excluded the zhmcclient_mock and tools directories from excluded section.

PR #393, opened 2017-08-10
Rename z Systems family to IBM Z

leopoldjuergen:
The z Systems family was rebranded to IBM Z.
This change renames documentation and code
in the project.


andy-maier:
We should also do that for zhmc-ansible-modules.

PR #411, opened 2017-08-30
[WIP] Added info about datetime limits to make output.

andy-maier:


andy-maier:
Abandoning this PR; it was just for trying this out on the range of CI platforms we use.

PR #432, opened 2017-09-13
Add CLI support for mount and unmount an ISO

leopoldjuergen:
With the new CLI command 'zhmc partition mountiso'
an user provided ISO image can be uploaded.
The uploaded ISO image is the image that gets attached
to the partition. If an ISO is attached to a partition,
it can be shown by the partition property 'boot-iso-image-name'.
To boot from this ISO image the partition property
'boot-device' must be set to 'iso-image' before starting
the partition.

With the new CLI command 'zhmc partition unmountiso'
the user unmounts the currently mounted ISO from
the partition. The partition property 'boot-iso-image-name'
is empty then.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -105,8 +105,8 @@ Released: not yet
 
 * Clarified in the description of `HbaManager.list()` that only the
   'element-uri' property is returned and can be used for filtering.
-  
-* The mock support for the "Create NIC" operation now performs more 
+
+* The mock support for the "Create NIC" operation now performs more
   checking on the URIs specified in the 'network-adapter-port' or
   'virtual-switch-uri' input properties, raising HTTP status 404 (Not Found)
   as specified in the HMC API book.
@@ -145,6 +145,8 @@ Released: not yet
 
 * Improved the error information in the ``ParseError`` exception, by adding
   the "Content-Type" header in cases where that is interesting.
+
+* Add CLI commmands to mount and unmount an ISO to a Partition.
 
 **Known issues:**
 
--- a/zhmccli/_cmd_partition.py
+++ b/zhmccli/_cmd_partition.py
@@ -15,6 +15,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import os
+
 import logging
 import click
 
@@ -439,6 +441,45 @@ def partition_console(cmd_ctx, cpc, partition, **options):
                                                       options))
 
 
+@partition_group.command('mountiso', options_metavar=COMMAND_OPTIONS_METAVAR)
+@click.argument('CPC', type=str, metavar='CPC')
+@click.argument('PARTITION', type=str, metavar='PARTITION')
+@click.option('--imagefile', type=str, required=True,
+              help='The file path of the ISO imagei file.')
+@click.option('--imageinsfile', type=str, required=True,
+              help='The file path of the INS file (within the file system '
+              'of the ISO image file).')
+@click.option('--boot', '-b', is_flag=True, required=False,
+              help='Set boot-device property to iso-image.')
+@click.pass_obj
+def partition_mount_iso(cmd_ctx, cpc, partition, **options):
+    """
+    Mount an ISO image to a partition.
+
+    In addition to the command-specific options shown in this help text, the
+    general options (see 'zhmc --help') can also be specified right after the
+    'zhmc' command name.
+    """
+    cmd_ctx.execute_cmd(lambda: cmd_partition_mount_iso(cmd_ctx, cpc,
+                                                        partition, options))
+
+
+@partition_group.command('unmountiso', options_metavar=COMMAND_OPTIONS_METAVAR)
+@click.argument('CPC', type=str, metavar='CPC')
+@click.argument('PARTITION', type=str, metavar='PARTITION')
+@click.pass_obj
+def partition_unmount_iso(cmd_ctx, cpc, partition):
+    """
+    Unmount an ISO image from a partition.
+
+    In addition to the command-specific options shown in this help text, the
+    general options (see 'zhmc --help') can also be specified right after the
+    'zhmc' command name.
+    """
+    cmd_ctx.execute_cmd(lambda: cmd_partition_unmount_iso(cmd_ctx, cpc,
+                                                          partition))
+
+
 def cmd_partition_list(cmd_ctx, cpc_name, options):
 
     client = zhmcclient.Client(cmd_ctx.session)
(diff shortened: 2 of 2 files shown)


zhmccli/_cmd_partition.py
```diff
@@ -439,6 +441,42 @@ def partition_console(cmd_ctx, cpc, partition, **options):
                                                       options))
 
 
+@partition_group.command('mountiso', options_metavar=COMMAND_OPTIONS_METAVAR)
+@click.argument('CPC', type=str, metavar='CPC')
+@click.argument('PARTITION', type=str, metavar='PARTITION')
+@click.option('--imagefile', type=str, required=True,
+              help='The path and name of the ISO image.')
```
andy-maier: Instead of "path and name of the ISO image", say "file path of the ISO image file".

zhmccli/_cmd_partition.py
```diff
@@ -439,6 +441,42 @@ def partition_console(cmd_ctx, cpc, partition, **options):
...
+@click.option('--imageinsfile', type=str, required=True,
+              help='The path and name of the ins file.')
```
andy-maier: Instead of "path and name of the ins file", say "file path of the INS file (within the file system of the ISO image file)".

zhmccli/_cmd_partition.py
```diff
@@ -439,6 +441,42 @@ def partition_console(cmd_ctx, cpc, partition, **options):
...
+@click.pass_obj
+def partition_mount_iso(cmd_ctx, cpc, partition, **options):
+    """
+    Mount an ISO image to a partition.
+
```
andy-maier: Add: "The boot options of the partition will be set to boot from the ISO file."

zhmccli/_cmd_partition.py
```diff
@@ -439,6 +441,42 @@ def partition_console(cmd_ctx, cpc, partition, **options):
...
+                                                        partition, options))
+
+
+@partition_group.command('unmountiso', options_metavar=COMMAND_OPTIONS_METAVAR)
+@click.argument('CPC', type=str, metavar='CPC')
+@click.argument('PARTITION', type=str, metavar='PARTITION')
+@click.pass_obj
+def partition_unmount_iso(cmd_ctx, cpc, partition):
+    """
+    Unmount an ISO image from a partition.
```
andy-maier: Add: "The boot options of the partition will be set to not boot from anything."

zhmccli/_cmd_partition.py
```diff
@@ -703,3 +741,35 @@ def cmd_partition_console(cmd_ctx, cpc_name, partition_name, options):
...
+               (image_name, partition.name))
+
+
+def cmd_partition_unmount_iso(cmd_ctx, cpc_name, partition_name):
+
+    client = zhmcclient.Client(cmd_ctx.session)
+    partition = find_partition(cmd_ctx, client, cpc_name, partition_name)
+
+    partition.pull_full_properties()
+    image_name = partition.get_property('boot-iso-image-name')
```
andy-maier: The `boot-iso-image-name` property can very well be set, but the partition is set to boot from something else (e.g. mount ISO image, followed by property update of `boot-device` to boot from something else).

I suggest that we don't use just a non-empty `boot-iso-image-name` property as an indicator that boot from ISO is set, but in addition (!) that `boot-device=='iso-image'`.

zhmccli/_cmd_partition.py
leopoldjuergen: Done

zhmccli/_cmd_partition.py
leopoldjuergen: Done

zhmccli/_cmd_partition.py
leopoldjuergen: This can be set by specifying the new '--boot'/'-b' option.
'

docs/changes.rst
```diff
@@ -56,6 +56,7 @@ Released: not yet
 * Avoided `DeprecationWarning` on Python 3 for invalid escape sequences
   in some places.
 
+<<<<<<< HEAD
```
andy-maier: A merge marker is in the `change.rst` file.

zhmccli/_cmd_partition.py
andy-maier: Ok.

zhmccli/_cmd_partition.py
andy-maier: Withdrawing this comment, because we now use specific options to set/unset the `boot-device` property.

zhmccli/_cmd_partition.py
andy-maier: Withdrawing this comment, because we now use specific options to set/unset the boot-device property.

zhmccli/_cmd_partition.py
```diff
@@ -703,3 +744,36 @@ def cmd_partition_console(cmd_ctx, cpc_name, partition_name, options):
...
+    if image_name:
+        partition.unmount_iso_image()
```
andy-maier: I suggest that the unmount command also gets a `--boot` option which sets the `boot-device` property to `'none'`.

andy-maier:
I have responded to my comments, and added some more comments, which all can be seen when looking at the file differences.

zhmccli/_cmd_partition.py
```diff
@@ -439,6 +441,47 @@ def partition_console(cmd_ctx, cpc, partition, **options):
...
+    """
+    cmd_ctx.execute_cmd(lambda: cmd_partition_mount_iso(cmd_ctx, cpc,
+                                                        partition, options))
+
+
+@partition_group.command('unmountiso', options_metavar=COMMAND_OPTIONS_METAVAR)
+@click.argument('CPC', type=str, metavar='CPC')
+@click.argument('PARTITION', type=str, metavar='PARTITION')
+@click.option('--boot', '-b', is_flag=True, required=False,
+              help='Unet boot-device property.')
```
andy-maier: Typo: "Unet" -> "Unset".

zhmccli/_cmd_partition.py
```diff
@@ -703,3 +746,38 @@ def cmd_partition_console(cmd_ctx, cpc_name, partition_name, options):
...
+
+def cmd_partition_unmount_iso(cmd_ctx, cpc_name, partition_name, options):
+
+    client = zhmcclient.Client(cmd_ctx.session)
+    partition = find_partition(cmd_ctx, client, cpc_name, partition_name)
+
+    partition.pull_full_properties()
+    image_name = partition.get_property('boot-iso-image-name')
+    if options['boot']:
+        partition.update_properties({'boot-device': ''})
```
andy-maier: Empty string is not a legal value for the `boot-device` property. Set it to the string `none`, instead.

andy-maier: Move the unsetting of the `boot-device` property after the unmounting, in case that one fails.

zhmccli/_cmd_partition.py
andy-maier: DONE

PR #451, opened 2017-09-26
Changed operator in constraints file from === to ==.

andy-maier:
**Note:** This PR is on top of PR #450 and targets 0.19.0!!
Ready for review and merge (after 0.18 has been released).


andy-maier:
Abandoned, because a reasonable use of `===` is to ensure that packages with local version identifiers are not used.

PR #469, opened 2017-11-29
Fixes #470: Added function test concept

### LINKED ISSUE #470: Debug logging enabled
### Actual behavior

PR #469 introduces function tests, including the ability to log to stderr. The issue is that debug logging is enabled even when not turned on (e.g. when using the default of "all=warning").

The issue shows up only when running the function tests introduced in PR #469 against a real HMC.

### Expected behavior

Debug logging is enabled only when turned on.

### Execution environment

* zhmcclient version: 0.18.0.dev
* Operating system (type+version): RHEL 7.3
* HMC version: 2.13.1
* CPC version: 2.13.1

andy-maier:
Ready for review and merge.

By default (e.g. in a CI), the function tests will either be skipped, or run against a faked/mocked environment. To set up testing against a real HMC/CPC, read the respective chapters in section "Development" in the docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,9 @@ test_py_files := \
     $(wildcard $(test_dir)/*/*.py) \
     $(wildcard $(test_dir)/*/*/*.py) \
 
+# Determine whether py.test has the --no-print-logs option.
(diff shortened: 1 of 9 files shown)


andy-maier:
Note that for some reason yet unknown, debug logging is on. We do not turn it on (at least I did not find where we would do that). It is turned on since this PR only.

We need to find out why that happens and get it turned off.

PR #526, opened 2018-09-05
Inspection of firmware features

andy-maier:
This PR will be used by the DPM storage group support.
Ready for review and merge.

**Important review point:**
* In case the feature is unknown or in case the CPC does not support features at all, should the `feature_enabled()` functions (1) raise `ValueError` as they do currently, or (2) return `False`? Returning `False`is more convenient, but does not allow distinguishing between feature available+disabled and feature unavailable (=unknown).


andy-maier:
Merged in agreement with reviewer without further review due to lack of time.

PR #534, opened 2018-09-17
Support more LPAR functionalities

mhartmay:
With these changes, the Tessia base library (https://gitlab.com/tessia-project/tessia-baselib) should be able to replace its self-written HMC library with this library.


zhmcclient/_lpar.py
```diff
@@ -790,6 +790,104 @@ def stop(self, wait_for_completion=True, operation_timeout=None,
...
+            :exc:`~zhmcclient.StatusTimeout` is raised.
+
+          allow_status_exceptions (bool):
+            Boolean controlling whether LPAR status "exceptions" is considered
+            an additional acceptable end status when `wait_for_completion` is
+            set.
+
+          force (bool):
+            Boolean controlling whether this operation is permitted when the
+            LPAR is in the "operating" status.
```
andy-maier: I suggest to move the "force" argument to the begin of the argument list. It is one that is put into the request body, and those we usually have at the begin of the list.

zhmcclient/_lpar.py
```diff
@@ -427,6 +427,150 @@ def deactivate(self, wait_for_completion=True,
...
+            Hexadecimal logical unit number (LUN) to be used for the
+            SCSI Load.
+
+          load_parameter (:term:`string`):
+            Optional load control string.  If empty string or `None`,
+            it is not passed to the HMC.
+
+          disk_partition_id (:term:`integer`):
+             Optional disk-partition-id (also called the boot program
+             selector) to be used for the SCSI Load.
```
andy-maier: Please state what the default is, if not specified. (The WS API book says the default is 0).

zhmcclient/_lpar.py
```diff
@@ -427,6 +427,150 @@ def deactivate(self, wait_for_completion=True,
...
+            Boolean controlling whether LPAR status "exceptions" is considered
+            an additional acceptable end status when `wait_for_completion` is
+            set.
+
+          force (bool):
+            Boolean controlling whether this operation is permitted when the
+            LPAR is in the "operating" status.
+
+            TBD: What will happen with the LPAR in that case (deactivated then
+            activated? nothing?)
```
andy-maier: I suggest to move the "force" argument to before the "wait_for_completion" argument. It is one that is put into the request body, and those we usually have before those that the zhmcclient adds.

andy-maier: I think the TBD can be removed in this case, because it is clear that the LPAR will IPL as a result.

zhmcclient/_lpar.py
```diff
@@ -500,6 +501,11 @@ def load(self, load_address=None, load_parameter=None,
             TBD: What will happen with the LPAR in that case (deactivated then
             activated? nothing?)
 
+          clear_indicator (bool):
+            Optional boolean controlling whether the memory should be
+            cleared before performing the load or not cleared. The
+            default value is `True`.
+
```
andy-maier: I suggest to move the "clear_indicator" argument to before the "wait_for_completion" argument. It is one that is put into the request body, and those we usually have before those that the zhmcclient adds.

zhmcclient/_lpar.py
mhartmay: Shall I also state what the default for force is?

zhmcclient/_lpar.py
mhartmay: Okay. I'll move "store_status_indicator" parameter as well.

zhmcclient/_lpar.py
mhartmay: Is the following description not better:  "If `None` it is not passed to the HMC."?

Because the defaults may change with newer HMC API versions.

mhartmay: In addition, I've changed the condition from `if disk_partition_id:` to `if disk_partition_id is not None:` because `bool(0) == False`.

PR #538, opened 2018-10-02
Support for specifiying ports

mhartmay:
Add support for specifying the used JMS/STOMP and HMC ports.

### DIFF
--- a/tests/unit/zhmcclient/test_notification.py
+++ b/tests/unit/zhmcclient/test_notification.py
@@ -37,7 +37,7 @@ class MockedStompConnection(object):
 
     def __init__(self, *args, **kwargs):
         """We ignore the args:
-            [(self._host, _STOMP_PORT)], use_ssl="SSL")
+            [(self._host, self._port)], use_ssl="SSL")
         """
         self._state_connected = False
         self._listener = None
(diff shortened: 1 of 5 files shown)


andy-maier:
@mhartmay The ports for the WS API and for JMS over STOMP are not configurable on the HMC. What is the purpose of making them configurable in the zhmcclient?

mhartmay:
> @mhartmay The ports for the WS API and for JMS over STOMP are not configurable on the HMC. What is the purpose of making them configurable in the zhmcclient?

Sometimes local port forwarding is needed, for example if you need SSH tunneling to access the HMC. When you're now trying to access multiple HMCs in parallel there is a port clash.

andy-maier:
Preliminary review: This PR introduces new optional `port` parameters to the `Session` and `NotificationReceiver` init methods. The implementation for that is fine.

**Discussion point:** The open discussion item is whether we want to go in that direction, as opposed to getting the port as part of the `host` parameter, using the standard URL syntax: `host:port`. This alternative would allow not having to change the higher levels of code sitting on top of the zhmcclient library (i.e. the zhmc CLI, and the zhmc Ansible modules, and the Ansible roles and playbooks using these modules). We'll discuss that in the zhmcclient team.

andy-maier:
I talked with several people and did not find that people have a strong preference between adding a new parameter vs. adding the port to the host string. Given that for IPv6 addresses, the URL notation is not backwards compatible to a hostname or IP address (in the URL there are square brackets around the IP address), I have a preference for the separate port parameter as implemented in the PR. Inthe layers above the client library, we can still start supporting URLs if we wanted to.

andy-maier:
I reviewed the code and it is fine. Merging.

mhartmay:
Thanks a lot!

PR #543, opened 2018-10-30
Update requests package to 2.20.0 in requirements.txt

leopoldjuergen:
The Requests package through 2.19.1 before 2018-09-14
for Python sends an HTTP Authorization header to
an http URI upon receiving a same-hostname
https-to-http redirect, which makes it easier
for remote attackers to discover credentials
by sniffing the network.

Link to vulnerability in National Vulnerability
Database:
https://nvd.nist.gov/vuln/detail/CVE-2018-18074

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -30,6 +30,10 @@ Released: not yet
 
 **Bug fixes:**
 
+* Update Requests package to 2.20.0 to fix follwing vulnerability of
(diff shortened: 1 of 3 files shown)


andy-maier:
Two comments:
1. In addition to fixing this in the master branch, this fix should also be rolled back into 0.20.1, and that version should then be released to Pypi.
2. Add an item in the change history about this.

Issue #549, opened 2018-12-21
FIltering by adapter-id returns empty result for hex digits


andy-maier:
### Actual behavior

Invoking:
```
adapter.find({'adapter-id': '11c'})
```
raises `NotFound` even when an adapter with that adapter-id exists.

This behavior is repeatably tied to the adapter-id containing hex digits ('a' to 'f'), and it happens for both upper case and lower case hex digits. 

### Expected behavior

The adapter should be found if it exists.

### Analysis

This turns out to be a defect in the Z firmware, which has been fixed and is now being rolled out. This issue should stay open until the fix is rolled out.

### Execution environment

* zhmcclient version: 0.21.0
* Operating system (type+version): MacOS
* HMC version: 2.14.0
* CPC version: 2.14.0

andy-maier:
Mitigated by PR #550.
Waiting for zFirmware fix to roll out; the fix will be in drivers D32L_HMC and D36C_HMC.

PR #550, opened 2018-12-21
Mitigated that fIltering by adapter-id returns empty result for hex digits

andy-maier:
Details:

* Added a mitigation for a firmware defect that causes filtering of
  adapters by adapter-id to return an empty result when the specified
  adapter-id contains hex digits ('a' to 'f'). See issue #549.

Signed-off-by: Andreas Maier <maiera@de.ibm.com>

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -32,6 +32,10 @@ Released: not yet
 
 **Enhancements:**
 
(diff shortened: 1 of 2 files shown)


andy-maier:
Pretty sure this works, based upon testing. Merging.

PR #577, opened 2019-05-10
Fixes #576: Changed exception raised when login user is not authorized

### LINKED ISSUE #576: Misleading ServerAuthError raised when creating user
### Actual behavior

Trying to create a User on the HMC with a login user that is not authorized to do that, results in the following exception being raised by the zhmcclient:
```
ServerAuthError: HTTP authentication failed: The API user does not have action/task permission to create a user of type standard
```

This is misleading, because it is not an authentication issue with the login user, but an authorization issue of the login user.

### Expected behavior

HTTPError is raised, instead.

### Execution environment

* zhmcclient version:
* HMC version:
* CPC version:

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -26,6 +26,18 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Operations that resulted in HTTP status 403, reason 1 ("The user under which
+  the API request was authenticated does not have the required authority to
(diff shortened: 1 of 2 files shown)


andy-maier:
Will change copyright for all files in an upcoming PR.

PR #587, opened 2019-08-29
Add more LPAR commands

leopoldjuergen:
Following LPAR commands are added:
- psw_restart()
- scsi_dump()

### DIFF
--- a/zhmcclient/_lpar.py
+++ b/zhmcclient/_lpar.py
@@ -1095,8 +1095,7 @@ class Lpar(BaseResource):
     def psw_restart(self, wait_for_completion=True, operation_timeout=None,
                     status_timeout=None, allow_status_exceptions=False):
         """
-        Restart the first available processor of this LPAR,
-        using the HMC operation "PSW Restart".
+        Restart this LPAR, using the HMC operation "PSW Restart".
 
         This HMC operation has deferred status behavior: If the asynchronous
         job on the HMC is complete, it takes a few seconds until the LPAR
--- a/zhmcclient/_lpar.py
+++ b/zhmcclient/_lpar.py
@@ -567,6 +567,141 @@ class Lpar(BaseResource):
         return result
 
     @logged_api_call
+    def scsi_dump(self, load_address, wwpn, lun, load_parameter=None,
+                  disk_partition_id=None,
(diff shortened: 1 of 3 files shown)


zhmcclient/_lpar.py
```diff
@@ -956,6 +1091,96 @@ def send_os_command(self, os_command_text, is_priority=False):
         self.manager.session.post(
             self.uri + '/operations/send-os-cmd', body)
 
+    @logged_api_call
+    def psw_restart(self, wait_for_completion=True, operation_timeout=None,
+                    status_timeout=None, allow_status_exceptions=False):
+        """
+        Restart the first available processor of this LPAR,
+        using the HMC operation "PSW Restart".
```
andy-maier: Talking about the "first available processor" is confusing, IMO. The effect is that the LPAR is restarted. Can we use the description of `Partition.psw_restart()`, also for consistency?

zhmcclient/_lpar.py
```diff
@@ -566,6 +566,141 @@ def scsi_load(self, load_address, wwpn, lun, load_parameter=None,
...
+        Load a standalone dump program from a designated SCSI device
+        in this LPAR, using the HMC operation "SCSI Dump".
+
+        This HMC operation has deferred status behavior: If the asynchronous
+        job on the HMC is complete, it takes a few seconds until the LPAR
+        status has reached the desired value. If `wait_for_completion=True`,
+        this method repeatedly checks the status of the LPAR after the HMC
+        operation has completed, and waits until the status is in the desired
+        state "operating", or if `allow_status_exceptions` was
+        set additionally in the state "exceptions".
```
andy-maier: An expected status makes only sense if the LPAR is restarted after the dump program has completed. Is that what happens? Unfortunately, the HMC API book is silent about that.

zhmcclient/_lpar.py
andy-maier: DONE - changed to "Restart this LPAR, using the HMC operation "PSW Restart"."

zhmcclient/_lpar.py
andy-maier: NO CHANGE.

PR #613, opened 2020-01-17
Changed Python versions for Travis CI

andy-maier:
See commit message.


andy-maier:
This change is included in PR #612, so this PR here can be closed.

PR #635, opened 2020-04-27
Start 0.26.1

andy-maier:
Note: Because the environment needed to be fixed first, this PR merges into master, and starts 0.26.1 from there. It also delivers change history updates that were forgotten.


andy-maier:
sigh - needs to target stable branch so it gets moved forward.

PR #653, opened 2020-08-10
Fixed installation from source dist archive; added installtest

andy-maier:
See commit message.


docs/intro.rst
```diff
@@ -249,6 +249,60 @@ target system from the downloaded packages:
...
+
+      # Install from the source dist archive:
+      $ tar -xf zhmcclient-0.26.1.tar.gz  # adjust version
+      $ python setup.py install
+
+* Pip install from repo root directory:
+
+  .. code-block:: text
+
+      $ git clone https://github.com/zhmcclient/python-zhmcclient
```
sabyadi, line 294: This can be done as one command like this:
```
pip install git+https://github.com/zhmcclient/python-zhmcclient
```

tests/installtest/test_install.sh
```diff
@@ -0,0 +1,500 @@
...
+PACKAGE_NAME=zhmcclient
+
+function abspath()
+{
+    # return absolute path name, normalizing ".."
+    # $1     : input path name
+    # return : normalized absolute path name
+    python -c "import os; print(os.path.abspath('$1'))"
+
+    # if [ -d "$1" ]; then
```
sabyadi, line 19: Can you please remove these comments?

docs/intro.rst
andy-maier: I know. I wanted to show this as separate statements, because the user might have a modified version of the repo, and with two commands they see more clearly what establishes the repo clone, and what performs the install from that repo.

tests/installtest/test_install.sh
andy-maier: DONE.

Issue #661, opened 2020-08-10
Address Pylint issues and enforce no Pylint issues except TODOs


andy-maier:

andy-maier:
A first chunk of issues is addressed in PR #662.

PR #662, opened 2020-08-10
First chunk of Pylint issue resolutions

andy-maier:
Partly fixes issue #661.

### DIFF
--- a/zhmcclient/_metrics.py
+++ b/zhmcclient/_metrics.py
@@ -344,6 +344,7 @@ class MetricsContext(BaseResource):
           :exc:`~zhmcclient.AuthError`
           :exc:`~zhmcclient.ConnectionError`
         """
+        # pylint: disable=protected-access
         self.manager.session.delete(self.uri)
         self.manager._metrics_contexts.remove(self)
 
@@ -408,6 +409,7 @@ class MetricDefinition(_MetricDefinitionTuple):
     """
 
     def __new__(cls, index, name, type, unit):
+        # pylint: disable=redefined-builtin
         """
         Parameters:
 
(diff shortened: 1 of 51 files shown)


zhmcclient/_metrics.py
```diff
@@ -906,7 +906,7 @@ def resource(self):
                 if found:
                     break
             else:
-                raise
+                raise AssertionError("No CPC found")
```
sabyadi, line 909: Is AssertionError the right exception type here? There are also three other loops that have a similar structure and do a raise without specifying the exception type in the else statement (resource_class == 'adapter', resource_class == 'partition' and resource_class == 'logical-partition'). So I would suggest to change them as well, since doing just raise outside of handling exception gives: 
```
Python 3.8.3 (default, Jul  8 2020, 14:27:55)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> raise
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: No active exception to reraise
>>>
```

andy-maier: It is not supposed to ever happen, so the assertion error indicates there is an error in the logic

andy-maier: And just having `raise` there is definitely incorrect, as you also have probven.
It just did not do any harm so far because it never happens :-)

andy-maier: Thanks for catching the other `raise` statements, I have added them back in.
Initially, I had fixed all of them but then I seem to have screwed up a rebase and forgot the others while fixing my screw-up.

PR #671, opened 2020-08-12
Fixed dependencies for development

andy-maier:
See commit message.

These changes partly come from experiences in other Python projects, which I am collecting in the https://github.com/andy-maier/cookiecutter-python repo..


andy-maier:
No, this is all manual. I maintain the cookiecutter-pythonproject meanwhile where some of that can be drawn, but for new issues that pop up it is a very manual process of finding out what the problem is. I only update the dependencies when there are issues, like some package removed support for Python 2.7 and things like that.

PR #672, opened 2020-08-12
Second chunk of Pylint issue resolutions that were not errors

andy-maier:
See commit message.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -88,13 +88,12 @@ sphinx-rtd-theme>=0.5.0
 # Pylint 1.x / astroid 1.x supports py27 and py34/35/36
 # Pylint 2.0 / astroid 2.0 removed py27, added py37
 # Pylint 2.4 / astroid 2.3 removed py34
-# TODO: Pinned pylint to <2.5.0 due to issues in 2.5.0 - remove in a while
 pylint>=1.6.4,<2.0.0; python_version == '2.7'
 pylint>=2.2.2,<2.4; python_version == '3.4'
-pylint>=2.4.4,<2.5.0; python_version >= '3.5'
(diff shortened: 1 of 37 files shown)


andy-maier:
The decrease in coverage surprised me and it turns out that pytest issues a lot of warnings:
```
/Users/maiera/virtualenvs/zhmc38/lib/python3.8/site-packages/_pytest/mark/structures.py:335
  /Users/maiera/virtualenvs/zhmc38/lib/python3.8/site-packages/_pytest/mark/structures.py:335: PytestCollectionWarning: cannot collect 'test_connectionerror_str_def' because it is not a function.
    def __call__(self, *args: object, **kwargs: object):  # noqa: F811
```
It seems that pytest does not support static test methods, unfortunately.
I will have to change that again.

andy-maier:
I have updated the PR to remove the `@staticmethod` decorator again from any test methods, and instead changed them to become global functions. Also, I removed the pinning of pylint to <2.5.0 because the issues in 2.5.0 have been addressed meanwhile.

vkahir:
It is interesting to know about pytest does not support static test methods. Thanks for the update Andy

mhartmay:
> I have updated the PR to remove the `@staticmethod` decorator again from any test methods, and instead changed them to become global functions. Also, I removed the pinning of pylint to <2.5.0 because the issues in 2.5.0 have been addressed meanwhile (and using 2.4.4 issues some useless warnings that 2.5.x does not issue).

Where did you find this information? At least what I've found is that staticmethods can be tested (see https://github.com/pytest-dev/pytest/pull/2531/files). Otherwise, this sounds like a bug in pytest. And to me the functions `create_session_1` and `create_session_2` look like they are candidates for a pytest fixture (see https://docs.pytest.org/en/stable/fixture.html).

PR #678, opened 2020-08-14
Third chunk of Pylint issue resolutions that were not errors

andy-maier:
See commit message.

### DIFF
--- a/tests/unit/zhmcclient_mock/test_idpool.py
+++ b/tests/unit/zhmcclient_mock/test_idpool.py
@@ -26,125 +26,152 @@ from zhmcclient_mock._idpool import IdPool
 urllib3.disable_warnings()
 
 
-class TestIdPool(object):
-    """All tests for class IdPool."""
+def test_init_error_1():
+    """
+    Test invalid IdPool.__init__() with lowest > highest.
+    """
 
-    def test_init_error_1(self):
+    with pytest.raises(ValueError):
+        IdPool(7, 6)
 
-        with pytest.raises(ValueError):
-            IdPool(7, 6)
 
-    def test_invalid_free_error_1(self):
+def test_invalid_free_error_1():
+    """
+    Test invalid IdPool.free() in some variations.
+    """
 
-        pool = IdPool(5, 5)
+    pool = IdPool(5, 5)
 
-        with pytest.raises(ValueError):
-            pool.free(4)  # not in range
+    with pytest.raises(ValueError):
+        pool.free(4)  # not in range
 
-        with pytest.raises(ValueError):
-            pool.free(5)  # in range but not allocated
+    with pytest.raises(ValueError):
+        pool.free(5)  # in range but not allocated
 
-        with pytest.raises(ValueError):
-            pool.free(6)  # not in range
+    with pytest.raises(ValueError):
+        pool.free(6)  # not in range
 
-    def test_invalid_free_error_2(self):
 
-        pool = IdPool(5, 5)
+def test_invalid_free_error_2():
+    """
+    Test invalid IdPool.free_if_allocated() in some variations.
+    """
 
-        pool.free_if_allocated(4)  # not in range (= not allocated)
+    pool = IdPool(5, 5)
 
-        pool.free_if_allocated(5)  # in range but not allocated
+    pool.free_if_allocated(4)  # not in range (= not allocated)
 
-        pool.free_if_allocated(6)  # not in range (= not allocated)
+    pool.free_if_allocated(5)  # in range but not allocated
 
-    def _test_exhausting_for_lo_hi(self, lowest, highest):
-        start = lowest
-        end = highest + 1
+    pool.free_if_allocated(6)  # not in range (= not allocated)
 
-        pool = IdPool(lowest, highest)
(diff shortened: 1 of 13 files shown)


vkahir (approved):

tests/unit/zhmcclient_mock/test_idpool.py
```diff
@@ -26,125 +26,156 @@
...
+    """
+    Internal helper function that tests exhausting an IdPool.
+    """
+
+    start = lowest
+    end = highest + 1
+
+    pool = IdPool(lowest, highest)
+
+    # Exhaust the pool
```
sabyadi, line 79: I suggest to replace the loop and set conversion with a simple dict like this:
```python
assert len({pool.alloc(): 1 for _ in range(start, end)}) == end - start
```

tests/unit/zhmcclient_mock/test_idpool.py
```diff
@@ -26,125 +26,156 @@
...
+    exhausting an IdPool.
+    """
+
+    start = lowest
+    end = highest + 1
+
+    pool = IdPool(lowest, highest)
+
+    # Exhaust the pool
+    id_list1 = []
```
sabyadi, line 106: I suggest to rewrite this as a list comprehension.

tests/unit/zhmcclient_mock/test_idpool.py
```diff
@@ -26,125 +26,156 @@
...
+
+    # Return everything to the pool
+    for _id in id_list1:
+        pool.free(_id)
+
+    # Verify that nothing is used in the pool
+    assert len(pool._used) == 0  # pylint: disable=protected-access
+
+    # Exhaust the pool
+    id_list2 = []
```
sabyadi, line 119: This as well.

tests/unit/zhmcclient_mock/test_idpool.py
```diff
@@ -26,125 +26,156 @@
...
+    """
+    Internal helper function that just combines _test_exhausting_for_lo_hi()
+    and _test_free_for_lo_hi().
+    """
+    # TODO: Convert to use fixtures or parametrization
+    _test_exhausting_for_lo_hi(lowest, highest)
+    _test_free_for_lo_hi(lowest, highest)
+
+
+def test_all():
```
sabyadi, line 142: How about we replace it with a parametrisation? Here is the code:
```python
@pytest.mark.parametrize("low,high", [
    (0, 0),
    (0, 1),
    (0, 9),
    (0, 10),
    (0, 11),
    (0, 19),
    (0, 20),
    (0, 21),
    (3, 3),
    (3, 4),
    (3, 9),
    (3, 10),
    (3, 11),
    (3, 12),
    (3, 13),
    (3, 14),
    (9, 9),
    (9, 10),
    (9, 11),
    (9, 18),
    (9, 19),
    (9, 20),
    (10, 10),
    (10, 11),
    (10, 19),
    (10, 20),
    (10, 21),
    (11, 11),
    (11, 12),
    (11, 20),
    (11, 21),
    (11, 22)])
def test_all(low, high):
    """
    Test pool exhaustion, etc. for some pool sizes.

    Knowing that the chunk size is 10, we focus on the sizes and range
    boundaries around that
    """
    _test_all_for_lo_hi(low, high)
```

tests/unit/zhmcclient_mock/test_urihandler.py
```diff
@@ -4668,6 +5766,7 @@ def test_invoke_ok(self):
         vswitch1['connected-vnic-uris'] = connected_nic_uris
 
         # the function to be tested:
+        # XXX: Fix this to be get instead of post, also in handler itself.
```
sabyadi, line 5769: And what does XXX stand for? :)

sabyadi (approved):
Please have a look at my comments.

tests/unit/zhmcclient_mock/test_idpool.py
andy-maier: I want the call that is being tested (in this case `pool.alloc()`) on a single line for multiple reasons: To set breakpoints, to step into it more easily, and to see in an exception traceback which statement failed.

tests/unit/zhmcclient_mock/test_idpool.py
andy-maier: See previous comment.

tests/unit/zhmcclient_mock/test_idpool.py
andy-maier: DONE - used parametrization as proposed, but with testcases in a constant to avoid their display when pytest detects a failure, and with the _test_all_for_lo_hi() helper function eliminated.

tests/unit/zhmcclient_mock/test_urihandler.py
andy-maier: Same as TODO and FIXME - it is reported by Pylint as a todo. I tend to use XXX for temporary TODOs that I want to address quickly, but in this case apparently I did not. :-(

andy-maier:
Adressed some of the comments, and responded to others.

PR #679, opened 2020-09-09
Added z15; Improved max partition detection

andy-maier:
See commit message.


andy-maier:
We need this fix in an internal project. Merging it without review.

vkahir (commented):
From my understanding with DPM design documents I have read recently max partitions update looks fine to me.

PR #706, opened 2020-12-21
Added partition.start_dump_program()

andy-maier:
See commit message.


andy-maier:
I'm merging this so it can started to be used by the zhmccli project.

PR #718, opened 2021-03-05
Improved logging and exc display for HTTP errors

andy-maier:
See commit message.


vkathir82 (approved):
HTTP Error changes looks good to me

PR #749, opened 2021-03-23
Added parameters to Lpar scsi methods; Fixed docs on default of Lpar scsi method

andy-maier:
See commit message.


andy-maier:
Merging without review.

PR #758, opened 2021-04-06
Fixed StatusTimeout when activating an LPAR

andy-maier:
Ready for test and review.


andy-maier:
@sean-mills tested it and confirms it works. Merging it.

Issue #762, opened 2021-04-07
Add support for a resource cache


andy-maier:
In support of https://github.com/zhmcclient/zhmc-prometheus-exporter/issues/112, it would be good to implement this at the level of the zhmcclient library by adding support for a resource cache.

The resource cache can be created by the user for a particular resource instance, and allows access to the properties of the resource. The properties are initially fetched via the corresponding GET request, and from that point on will be automatically updated by registering for property changes (which will be delivered back to the client as notifications).

Things to check:
* Which properties can change but do not have change notifications?

One possible solution:
* For each resource class (e.g. Partition) introduce a cached version (e.g. CachedPartition) that is initialized from the original (uncached) resource instance.
* Ideally, the cached class would be a subclass of the original class.
* The cached class has an enable() method that registers for property changes of the resource instance and performs a full fetch of properties.
* The cached class has a disable() method that unregisters for property changes.
* In the init method of the cached class performs enable()
* The notification handler looks up the cached resource by URL and updates its properties.

andy-maier:
PR #768 is a precondition for this since it ensures (for the most part) that users cannot modify the properties.

andy-maier:
Since the control of which resource class is instantiated is within the client and not on the side of the user, there needs to be some way to enable and disable caching.

PROPOSAL:
* The Client class stores the changing-enabled status for each type of resource, and has methods to control that.

Also, probably it is easier if the existing resource classes get the caching ability added rather than creating new classes.

PR #768, opened 2021-04-14
Changed properties attribute to be immutable

andy-maier:
See commit message.
This is a preparation for resource caching as proposed in issue #762, since it ensures that users cannot modify the properties dictionary.

**This is for review and discussion, not yet ready to merge.**


vkathir82 (approved):
The idea of immutable properties looks good to me and I feel this the good idea of having the cache copy updated for these additional values which are not supported by WS API metric data. If I think any other better/alternate implementation strategy, I will update here. Conceptually this idea looks good to me.

PR #769, opened 2021-04-16
Added server cert verification, security docs, trouble shooting docs

andy-maier:
See commit message.

Note that the new `verify_cert` parameter has a default of `False` in order to be backwards compatible. Thinking about that should be one review point.

Direct links for reviewers:

* New "Security" section: https://github.com/zhmcclient/python-zhmcclient/blob/andy/add-verify-cert/docs/security.rst
* New "Troublsshooting" section: https://github.com/zhmcclient/python-zhmcclient/blob/andy/add-verify-cert/docs/appendix.rst#troubleshooting
* New `verify_cert` init parameter: https://github.com/zhmcclient/python-zhmcclient/blob/andy/add-verify-cert/zhmcclient/_session.py#L357
 
Note that the rendered .RST files show some of the links in a strange way. The built documentation will have them right.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -22,6 +22,72 @@ This section contains information that is referenced from other sections,
 and that does not really need to be read in sequence.
 
 
+.. _`Troubleshooting`:
+
+Troubleshooting
+---------------
+
+This section describes a few issues and how to address them.
+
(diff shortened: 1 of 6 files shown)


vkathir82 (approved):
Changes looks good to me

PR #783, opened 2021-06-08
Improved exception handling for HMC certificate failures

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -59,6 +59,10 @@ Released: not yet
   SSL/TLS handshake. By default, the certificate is validated against
   the CA certificates provided in the Python 'certifi' package. (issue #779)
 
+* Added catching of OSError/IOError exceptions raised by the 'requests' package
+  for certain certificate validation failures, re-raising such exceptions as a
+  pywbem.ConnectionError.
+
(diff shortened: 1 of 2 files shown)


vkathir82 (commented):
@andy-maier I see that In Python3 they removed **IOError** and introduced **BlockingIOError** as a subclasses of **OSError** itself. As we support for both PY2 and PY3 environments now, will we get any issue with PY3 having **IOError** ?

https://docs.python.org/3/library/exceptions.html

andy-maier:
This change is for catching exceptions raised by requests methods. I just verified that again with a cert file that does not exist and thus causes this exception to be raised: The requests method raises IOError on Python 2 and OSError on Python 3.

On Python 3, IOError and OSError are the same class (same ID). On Python 2, they are unrelated classes (i.e. no subclass relationship). So the case I tested would also work on all Python versions if we just catch IOError in the code, but I'm not sure we can guarantee that OSError is not used on Python 2. So I think catching both puts us on the safe side.

vkathir82 (approved):
@andy-maier Thanks for the clarification of the exception usage. Changes are looking good to me

PR #790, opened 2021-06-17
Start 1.0.0

andy-maier:
No review needed.


andy-maier:
The 1.0.0 version that was started with this PR was lateron changed to become 0.32.0

PR #795, opened 2021-06-22
Added support for Console.list_permitted_partitions/_lpars()

andy-maier:
See commit message.


vkathir82 (approved):
Changes looks good with list partitions

PR #797, opened 2021-06-23
Fixed/added authorization reqs for Partition and Lpar methods

andy-maier:


andy-maier:
Only docs changes - merging without further review.

PR #834, opened 2021-10-05
Start 0.32.2

andy-maier:


andy-maier:
Abandoning this PR, since 0.32 is no longer maintained. 1.0 is the fix version, instead.

PR #855, opened 2021-11-09
Added Cpc.export_dpm_configuration() method

andy-maier:
This PR needs to wait for discussion with the DPM team before being merged.

Nevertheless, it is ready for review and serious use.


andy-maier:
Still searching for an error

andy-maier:
Error is fixed, the exported DPM config has now been verified against the plain inventory for most resource classes.

PR #910, opened 2022-02-09
Fixes in mock support for "Update LPAR Properties"

andy-maier:
See commit message for details.


andy-maier:
Since this is just the mock support and it works in the tests, I am merging this without review.

PR #1002, opened 2022-06-21
Fixed filepath AttributeError when using testutils with mocked envs

andy-maier:


andy-maier:
We need to release 1.3.1 - merging this PR without further review.

PR #1005, opened 2022-06-21
Fixed dumping of HMC definitions; Added dump_hmc_definition.py example

andy-maier:


andy-maier:
We need to release 1.3.1 - merging this PR without further review.

PR #1025, opened 2022-06-30
Release 1.3.2

andy-maier:
No review needed.


andy-maier:
Expected to fail due to coveralls.io maintenance. See https://status.coveralls.io/

PR #1056, opened 2022-10-02
Added support for auto-updated resource managers

andy-maier:


GitCop:

 * Commit: ecc7a45dab95fb8e588fe00e5631e63f5c9bec3c

---

examples/show_auto_updated_partition_manager.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python
+# Copyright 2022 IBM Corp. All Rights Reserved.
```
simon-spinner, line 2: We now have 2023

examples/show_auto_updated_partition_manager.py
```diff
@@ -0,0 +1,177 @@
...
+mode.
+
+The shorter durations of list() indicate that the locally maintained list was
+used; the longer durations indicate that a list operation was performed on the
+HMC.
+"""
+
+import sys
+import uuid
+# import logging
```
simon-spinner, line 27: Why is this import commented out?

examples/show_auto_updated_partition_manager.py
```diff
@@ -0,0 +1,177 @@
...
+    part_name = "zhmc_test_{}".format(uuid.uuid4())
+    print("Creating partition {} ...".format(part_name))
+    part_props = {
+        'name': part_name,
+        'description': 'Original partition description.',
+        'cp-processors': 2,
+        'initial-memory': 1024,
+        'maximum-memory': 2048,
+        'processor-mode': 'shared',
+        'boot-device': 'test-operating-system'
```
simon-spinner, line 112: We should not use 'test-operating-system' - it is not documented and not available on all machines.

andy-maier: This was just a rename of the script, but ok so let's fix it. I have removed the boot-device property as it is not needed to the purpose of showing property updates.

PR #1082, opened 2023-01-30
Fixed docs issues & missing datetime_from_timestamp/timestamp_from_datetime

andy-maier:
This fixes the recent weekend build issue on Python 3.10

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -208,7 +208,7 @@ Released: 2022-05-17
   (issues #950, #986)
 
 * Renamed the properties of the 'zhmcclient.testutils.HMCDefinition' to remove
-  the 'hmc_' prefix, e.g. 'hmc_userid' became 'userid', etc. (part of issue #986)
+  the 'hmc&nbsp;_' prefix, e.g. 'hmc_userid' became 'userid', etc. (part of issue #986)
 
 **Bug fixes:**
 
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -208,7 +208,7 @@ Released: 2022-05-17
   (issues #950, #986)
 
 * Renamed the properties of the 'zhmcclient.testutils.HMCDefinition' to remove
-  the 'hmc_' prefix, e.g. 'hmc_userid' became 'userid', etc. (part of issue #986)
+  the 'hmc&nbsp;_' prefix, e.g. 'hmc_userid' became 'userid', etc. (part of issue #986)
```
EdGue42, line 211: &nbsp; is on purpose, or is that a typo?

andy-maier: That is on purpose. `&nbsp;` is the HTML symbol for non-breaking space which means it does not create any visible character but represents a character for the parsing. So the result looks like `'hmc_'` but it is apparently no longer parsed by RST as a link (see the builddoc item in https://github.com/zhmcclient/python-zhmcclient/actions/runs/4044091687/jobs/6953821628)

Actually, if I think about it, I don't know why it would no longer do that, but it seems to work.

EdGue42 (approved):

PR #1155, opened 2023-05-05
Added support for Group resources

andy-maier:

Particular review points:
* Should the `groups` property providing the GroupManager object be in the Client object (as currently in this PR) or in the Console object?

### DIFF
--- a/tests/end2end/test_group.py
+++ b/tests/end2end/test_group.py
@@ -1,0 +1,156 @@
+# Copyright 2023 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+End2end tests for groups.
+
+These tests do not change any existing groups, but create, modify and delete
+groups for testing purposes.
+"""
+
+from __future__ import absolute_import, print_function
+
+import warnings
+import pytest
+from requests.packages import urllib3
+
+import zhmcclient
+# pylint: disable=line-too-long,unused-import
+from zhmcclient.testutils import hmc_definition, hmc_session  # noqa: F401, E501
+# pylint: enable=line-too-long,unused-import
+
+from .utils import pick_test_resources, runtest_find_list, TEST_PREFIX, \
+    skip_warn, skipif_no_group_support
+
+urllib3.disable_warnings()
+
+# Properties in minimalistic PasswordRule objects (e.g. find_by_name())
+GROUP_MINIMAL_PROPS = ['object-uri', 'name']
+
+# Properties in PasswordRule objects returned by list() without full props
+GROUP_LIST_PROPS = ['object-uri', 'name']
+
+# Properties whose values can change between retrievals of PasswordRule objects
+GROUP_VOLATILE_PROPS = []
+
+
+def test_group_find_list(hmc_session):  # noqa: F811
+    # pylint: disable=redefined-outer-name
+    """
+    Test list(), find(), findall().
+    """
+    client = zhmcclient.Client(hmc_session)
+    console = client.consoles.console
+    hd = hmc_session.hmc_definition
+
+    skipif_no_group_support(client)
+
+    # Pick the groups to test with
+    group_list = console.groups.list()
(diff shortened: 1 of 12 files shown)


tests/end2end/test_group.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Test list(), find(), findall().
+    """
+    client = zhmcclient.Client(hmc_session)
+    hd = hmc_session.hmc_definition
+
+    api_version = client.query_api_version()
+    hmc_version = api_version['hmc-version']
+    hmc_version_info = tuple(map(int, hmc_version.split('.')))
+    if hmc_version_info < (2, 13, 0):
```
EdGue42, line 59: could that go into a utility method?

tests/end2end/test_group.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Test create, read, update and delete a group.
+    """
+    client = zhmcclient.Client(hmc_session)
+    hd = hmc_session.hmc_definition
+
+    api_version = client.query_api_version()
+    hmc_version = api_version['hmc-version']
+    hmc_version_info = tuple(map(int, hmc_version.split('.')))
+    if hmc_version_info < (2, 13, 0):
```
EdGue42, line 88: Have a look into the internal github, I added a skipif ... function to the utility code already.
Helps with code duplication.

tests/unit/zhmcclient/test_group.py
```diff
@@ -0,0 +1,261 @@
...
+    def test_group_delete(self, input_props, exp_exc):
+        """Test Group.delete()."""
+
+        faked_group = self.add_group(name=input_props['name'])
+
+        group_mgr = self.client.groups
+        group = group_mgr.find(name=faked_group.name)
+
+        if exp_exc is not None:
+
```
EdGue42, line 208: maybe that is personal choice, but I am surprised how many newlines you put in your code.

simon-spinner (commented):
So far we put all resources below cpcs/consoles. What makes a group different so that it should be part of the client itself?

tests/end2end/test_group.py
```diff
@@ -0,0 +1,165 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+End2end tests for groups.
+
+These tests do not change any existing groups, but create,
+modify and delete test groupus.
```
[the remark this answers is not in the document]
andy-maier: No - the Atom editor does not do that :-)
Fixed.

tests/end2end/test_group.py
andy-maier: Good point. 
I moved these lines into a new function `skipif_no_group_support()`, similar to other such functions.

tests/end2end/test_group.py
```diff
@@ -0,0 +1,165 @@
...
+                  format(h=hd.host, v=hmc_version))
+
+    # Pick the groups to test with
+    group_list = client.groups.list()
+    if not group_list:
+        skip_warn("No groups defined on HMC {h}".format(h=hd.host))
+    group_list = pick_test_resources(group_list)
+
+    for group in group_list:
+        print("Testing with group {r!r}".format(r=group.name))
```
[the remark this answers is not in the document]
andy-maier: Because it was copied from the password rule support ("r"=rule).
Changed to "g".

tests/end2end/test_group.py
andy-maier: DONE

tests/end2end/test_group.py
```diff
@@ -0,0 +1,165 @@
...
+        'match-info': {},
+    }
+    group_auto_props = {}
+
+    try:
+
+        # The code to be tested
+        try:
+            group = client.groups.create(
+                group_input_props)
```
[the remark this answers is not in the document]
andy-maier: Agreed. Removed the line break.

andy-maier:
@simon-spinner Whether to put Group objects under Client or under Console: That is a good question. I did not have a particular reason other than the URI structure. But then we have put some resources under Console such as UserRole that also have top-level URIs.
Also, what puzzled me recently is that for BCPII, the SE is the console.
Is there any guidance from the HMC team?

simon-spinner:
> @simon-spinner Whether to put Group objects under Client or under Console: That is a good question. I did not have a particular reason other than the URI structure. But then we have put some resources under Console such as UserRole that also have top-level URIs. Also, what puzzled me recently is that for BCPII, the SE is the console. Is there any guidance from the HMC team?

The UserRole is a managed object and as such has a top-level URI. However the list operation is below the console. For some other objects, including the Group object, the list operation is top-level as well. Not sure what the reasons for that are. The Group object surprised me because it is local to a certain HMC.

simon-spinner (approved):

PR #1161, opened 2023-05-07
timeout added to allow hmc to wait for partition to stop

andy-maier:
method: stop() parameter: status_timeout

updated enhancements to include status timeout for partition stop


andy-maier:
This change was already integrated with PR #699

PR #1177, opened 2023-05-13
Added list_sibling_adapters() method to Adapter class

andy-maier:

Particular review points:
* Is "sibling" an appropriate term for the other adapters on the same card

### DIFF
--- a/tests/end2end/test_adapter.py
+++ b/tests/end2end/test_adapter.py
@@ -467,3 +467,100 @@ def test_adapter_list_assigned_part(dpm_mode_cpcs):  # noqa: F811
             # Cleanup
             if tmp_part:
                 tmp_part.delete()
+
+
+def base_adapter_id(adapter_id, family):
+    """
+    Return the base adapter_id of an adapter_id.
+    Both are in hex.
+    """
+
+    # Adapter families that have no physical adapter card
+    virtual_families = [
+        'hipersocket',
+        'coupling',  # Internal Coupling Facility
+        'ism',  # Internal Shared Memory
+    ]
+
+    pchid = int(adapter_id, 16)
+
+    if family in virtual_families:
+
+        # Make sure the virtual card detection based on the PCHID range in
+        # Adapter.list_sibling_adapters() holds true
+        assert pchid >= int('7c0', 16)
+
+        # Virtual adapters have only 1 PCHID
+        base_pchid = pchid
+
+    else:
+
+        # Make sure the physical card detection based on the PCHID range in
+        # Adapter.list_sibling_adapters() holds true
+        assert pchid < int('7c0', 16)
+
+        # Physical adapters have 4 PCHIDs reeserved for the slot.
+        # Calculate the base PCHID for the slot.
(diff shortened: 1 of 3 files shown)


tests/end2end/test_adapter.py
```diff
@@ -467,3 +467,79 @@ def test_adapter_list_assigned_part(dpm_mode_cpcs):  # noqa: F811
...
+def base_adapter_id(adapter_id, family):
+    """
+    Return the base adapter_id of an adapter_id. Both are in hex.
+    """
+    # Adapter families that have no physical adapter card
+    logical_families = [
+        'hipersocket',
+    ]
+    pchid = int(adapter_id, 16)
+    if family in logical_families:
```
EdGue42, line 481: your if assigns the same value for then/else ?

tests/end2end/test_adapter.py
```diff
@@ -467,3 +467,79 @@ def test_adapter_list_assigned_part(dpm_mode_cpcs):  # noqa: F811
...
                 tmp_part.delete()
+
+
+def base_adapter_id(adapter_id, family):
+    """
+    Return the base adapter_id of an adapter_id. Both are in hex.
+    """
+    # Adapter families that have no physical adapter card
+    logical_families = [
+        'hipersocket',
```
simon-spinner, line 478: What about ism and icp?

zhmcclient/_adapter.py
```diff
@@ -679,3 +679,58 @@ def list_assigned_partitions(self, full_properties=False, filter_args=None):
...
+          :exc:`~zhmcclient.AuthError`
+          :exc:`~zhmcclient.ConnectionError`
+        """
+        # This algorithm is based on the fact that physical adapter cards
+        # have their PCHIDs always within a range of 4 adjacent PCHIDs that
+        # start at a multiple of 4.
+
+        self_pchid = int(self.prop('adapter-id'), 16)
+        if self_pchid >= int('700', 16):
+            # Hipersocket adapter
```
simon-spinner, line 726: just on your comment: could be also an ISM or ICP adapter (they also have VCHIDs >700).

tests/end2end/test_adapter.py
andy-maier: The code says:
```
    if family in logical_families:
        base_pchid = pchid
    else:
        base_pchid = pchid // 4 * 4
```
That is not the same value.

zhmcclient/_adapter.py
andy-maier: With this code, I want to detect adapters without siblings, where sibling is defined as the set of other PCHIDs on the same same physical card. So I need to understand more details about ISM and ICP.

andy-maier: Changed that to "Virtual adapter" and added more explanatory comments.

tests/end2end/test_adapter.py
andy-maier: Added "ism" and "coupling" to the list of virtual families and renamed the variable to "virtual_families".

PR #1191, opened 2023-07-03
Circumvented the removal of Python 2.7 from setup-python

andy-maier:


andy-maier:
The tests fail due to new safety issues. Will be fixed in a later PR.

PR #1227, opened 2023-07-30
Test: Fixtures for HTTP-level mocking

andy-maier:
This PR provides the basis for multiple other PRs that add testcases based on HTTP-level mocking.

### DIFF
--- a/tests/common/http_mocked_fixtures.py
+++ b/tests/common/http_mocked_fixtures.py
@@ -1,0 +1,203 @@
+# Copyright 2023 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tests/common/http_mocked_fixtures.py
```diff
@@ -0,0 +1,203 @@
...
+
+    A test function parameter using this fixture resolves to a
+    :class:`~zhmcclient.Cpc` object that is in DPM mode but has only a minimal
+    set of properties in the object (those that are returned with the 'List
+    CPCs' operation).
+    """
+
+    client = zhmcclient.Client(http_mocked_session)
+
+    with requests_mock.mock() as m:
```
EdGue42, line 106: Would it make sense that the values you are passing here would be arguments (with defaults) to this function?

tests/common/http_mocked_fixtures.py
```diff
@@ -0,0 +1,203 @@
...
+        cpcs = client.cpcs.list()
+        cpc = cpcs[0]
+
+    return cpc
+
+
+@pytest.fixture(
+    scope='module'
+)
+def http_mocked_cpc_classic(request, http_mocked_session):  # noqa: F811
```
EdGue42, line 128: if you could pass that the flag as parameter, you wouldn't need two functions.

andy-maier: You cannot have pytest fixtures that take additional arguments. Pytest fixtures are resolved by looking for a matching parameter name on the test function.

PR #1241, opened 2023-08-31
Support for requesting additional properties in list() methods

andy-maier:
This is needed for https://github.com/zhmcclient/zhmc-ansible-modules/issues/651

### DIFF
--- a/zhmcclient/_manager.py
+++ b/zhmcclient/_manager.py
@@ -36,7 +36,8 @@ from nocasedict import NocaseDict
 
 from ._logging import logged_api_call
 from ._exceptions import NotFound, NoUniqueMatch, HTTPError
-from ._utils import repr_list
+from ._utils import repr_list, matches_filters, divide_filter_args, \
+    make_query_str, RC_LOGICAL_PARTITION
 
 __all__ = ['BaseManager']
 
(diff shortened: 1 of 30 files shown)


zhmcclient/_manager.py
```diff
@@ -672,6 +673,200 @@ def supports_properties(self):
...
+                        ','.join(additional_properties))
+                    query_parms.append(ap_parm)
+                query_parms_str = make_query_str(query_parms)
+                uri = '{}{}'.format(list_uri, query_parms_str)
+
+                try:
+                    result = self.session.get(uri)
+                except HTTPError as exc:
+                    if exc.http_status == 404 and exc.reason == 1:
+                        # Unlike other list operations, "List Logical Partitions
```
EdGue42, line 763: 404.1 is only expected for a specific case; maybe you want to check explicitly for that case, to avoid ignoring 404.1 on other operations?

zhmcclient/_manager.py
```diff
@@ -672,6 +673,200 @@ def supports_properties(self):
...
+                        # of CPC" fails with 404.1 "ERROR: found no Images" if
+                        # no LPAR matches the filters in the query parms.
+                        result = []
+
+                if result:
+                    props_list = result[result_prop]
+                    for props in props_list:
+
```
EdGue42, line 771: minor: remove empty line?

zhmcclient/_manager.py
```diff
@@ -672,6 +673,200 @@ def supports_properties(self):
...
+        Raises:
+
+          :exc:`~zhmcclient.HTTPError`
+          :exc:`~zhmcclient.ParseError`
+          :exc:`~zhmcclient.AuthError`
+          :exc:`~zhmcclient.ConnectionError`
+        """
+        resource_obj_list = []
+        if self.auto_update_enabled() and not self.auto_update_needs_pull():
```
EdGue42, line 845: code duplication: the above list() function contains the same if/then?

EdGue42 (commented):
As usual, just commenting for now ... one big question: you enhanced a lot of existing list methods. Obviously, your existing tests will cover the existing functionality, but wouldn't you want to have (at least a few) tests covering the "changed" behavior of the list methods?

zhmcclient/_manager.py
andy-maier: Probably that would be good - checked via `isinstance(self, LparManager)`.

In addition, we need to re-raise the exception in all other cases, that was also missing. Fixed.

zhmcclient/_manager.py
andy-maier: DONE

andy-maier:
I agree to the need for tests. 
I tried to add tests for additional_properties, but it turns out that PR #1251 is needed for that. Instead of creatng even more depenencies between the PRs (it is already pretty bad with some of these 5 PRs), I have created issue #1252 to cover that.

EdGue42 (approved):

PR #1245, opened 2023-09-01
Docs: Improved documentation for developing tests

andy-maier:
To look at the section that has been updated:
```
git checkout andy/testutils-docs
make builddoc
open build_doc/html/docs/development.html
```
In the browser page that opens, navigate to section 8.6 "Developing tests".

### DIFF
--- a/zhmcclient/testutils/_cpc_fixtures.py
+++ b/zhmcclient/testutils/_cpc_fixtures.py
@@ -57,9 +57,12 @@ def all_cpcs(request, hmc_session):  # noqa: F811
     The CPCs to test against are defined in the ``cpcs`` variable for the
     HMC entry in the :ref:`HMC inventory file`.
 
-    A test function parameter using this fixture resolves to a list of
-    :class:`zhmcclient.Cpc` objects representing that set of CPCs. These
+    A test function parameter with the name of this fixture resolves to a list
+    of :class:`zhmcclient.Cpc` objects representing that set of CPCs. These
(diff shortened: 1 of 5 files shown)


zhmcclient/testutils/_cpc_fixtures.py
```diff
@@ -57,10 +57,12 @@ def all_cpcs(request, hmc_session):  # noqa: F811
     The CPCs to test against are defined in the ``cpcs`` variable for the
     HMC entry in the :ref:`HMC inventory file`.
 
-    A test function parameter using this fixture resolves to a list of
-    :class:`zhmcclient.Cpc` objects representing that set of CPCs. These
+    A test function parameter with the name of this fixture resolves to a list
+    of :class:`zhmcclient.Cpc` objects representing that set of CPCs. These
     objects have the "short" set of properties from :meth:`zhmcclient.Cpc.list`.
 
+    The test function is repeated for each CPC in that set of CPCs.
```
EdGue42, line 64: Well, this doesn't happen automatically, right? 
The developer has to write the corresponding code, so this statement here is more like guidance how tests should be written?! If so, that could be made more clear.

andy-maier: That is a good catch - copy & paste error. In this case, the fixture returns a list, and the test function needs to iterate through the list (or pick one). Will fix it.

andy-maier: DONE

EdGue42 (approved):

anilak11 (approved):

PR #1246, opened 2023-09-02
Fixed end2end tests for example mocked environment

andy-maier:

With this PR, the set of errors when running against the example mock support has been reduced to these, which are caused by not having implemented support for the corresponding operations in the mock support. This is left for another PR to add.

```
$ TESTINVENTORY=examples/example_hmc_inventory.yaml TESTVAULT=examples/example_hmc_vault.yaml make end2end
. . .
FAILED tests/end2end/test_adapter.py::test_adapter_list_assigned_part[hmc_definition=MOCKED_Z16_DPM] - zhmcclient._exceptions.HTTPError: 404,1: Unknown resource with URI: /api/adapters/osa1/operations/get-partitions-assigned-to-adapter [GET /api/adapters/osa1/operations...
FAILED tests/end2end/test_cpc.py::test_cpc_export_dpm_config[hmc_definition=MOCKED_Z16_DPM] - zhmcclient._exceptions.HTTPError: 404,1: Unknown resource with URI: /api/services/inventory [POST /api/services/inventory]
```

### DIFF
--- a/zhmcclient_mock/_urihandler.py
+++ b/zhmcclient_mock/_urihandler.py
@@ -213,22 +213,31 @@ class ServerError(HTTPError):
             message=message)
 
 
-def parse_query_parms(method, uri, query_str):
+def parse_query_parms(method, uri):
     """
-    Parse the specified query parms string and return a dictionary of query
-    parameters. The key of each dict item is the query parameter name, and the
+    Parse the specified URI into URI without query parms and query parms,
+    and return a dictionary of query parameters.
+
+    The key of each dict item is the query parameter name, and the
     value of each dict item is the query parameter value. If a query parameter
     shows up more than once, the resulting dict item value is a list of all
     those values.
 
-    query_str is the query string from the URL, everything after the '?'. If
-    it is empty or None, None is returned.
-
     If a query parameter is not of the format "name=value", an HTTPError 400,1
     is raised.
+
+    Returns:
+        tuple(uri, query_parms) with:
+        - uri(str): Input URI without query parms
+        - query_parms(dict): Query parms as dict (name, value). Empty if no
+          query parms were specified.
     """
-    if not query_str:
-        return None
+    uri_split = uri.split('?')
+    uri = uri_split[0]
+    try:
+        query_str = uri_split[1]
+    except IndexError:
+        query_str = ''
     query_parms = {}
     for query_item in query_str.split('&'):
         # Example for these items: 'name=a%20b'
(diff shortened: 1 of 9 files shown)


GitCop:

 * Commit: 795426584448e1e492872c12df5e15b0c8ec72e2

---

zhmcclient_mock/_urihandler.py
```diff
@@ -387,6 +396,23 @@ def check_set_noninput(method, uri, properties, prop_name, prop_value):
...
 
+def check_invalid_query_parms(method, uri, query_parms, valid_query_parms):
+    """
+    Check that the query parameters are valid.
+
+    Raises:
+      BadRequestError with reason 1.
+    """
+    for qp in query_parms:
+        if qp not in valid_query_parms:
```
EdGue42, line 407: Would it make sense to collect all invalid query params, and throw the exception after the loop, in case the set (string) with invalid query params isn't empty?
Why tell the user only about one problem at a time, when you could report all of them at once?

zhmcclient_mock/_urihandler.py
```diff
@@ -453,17 +479,37 @@ class GenericGetPropertiesHandler(object):
...
             new_exc.__cause__ = None
             raise new_exc  # zhmcclient_mock.InvalidResourceError
-        return resource.properties
+
+        subset_pnames = query_parms.get('properties', None)
+        if subset_pnames:
+            subset_pnames = subset_pnames.split(',')
+            ret_props = {}
+            for pname in subset_pnames:
+                ret_props[pname] = resource.properties[pname]
```
EdGue42, line 510: I assume: if one uses an invalid property name, you end up with a python key error here?
Is there a way to avoid that (if we would want to avoid it)

zhmcclient_mock/_urihandler.py
andy-maier: DONE.

zhmcclient_mock/_urihandler.py
andy-maier: If the property specified in the query parms is invalid, it has been detected before in `check_invalid_query_parms()`.

The other question is whether the resource has the property available. `hmc.lookup_by_uri()` returns the FakedResource, i.e. the internal data representation of the resource. It has all properties that have been defined when the faked resource was created. So there is a chance that the creation of the resource missed the (valid) property. I defined a new exception class `MockedResourceError` for this kind of error and added a check for the presence of the property.

EdGue42 (approved):

EdGue42:
Where: I do think I saw this one problem that you resolved here (about collecting all invalid params and adding all those to the exception) in some other piece of code I looked at today. 🤪

andy-maier:
The other place was probably in `parse_query_parms()` with the check for valid name=value formats.

Keep in mind: This is just the mock support, which is used only for testing, and it is much more simple minded than the HMC. For example, it does all checks just upon use, and not upfront like the HMC probably does. I think we have to live with compromises for the error returns of the mock support.

PR #1256, opened 2023-09-07
Make target 'end2end_mocked' to run end2end tests against example mocked env

andy-maier:


simon-spinner (approved):

EdGue42 (commented):
Thats a neat thing to be able to smoke test with end2end tests!

PR #1266, opened 2023-09-15
Fixed max number of threads in bulk ops

andy-maier:


EdGue42 (approved):
Looks okay to me, albeit I couldn't find that information in the API book.

andy-maier:
In the type foeld of the threads parameter, it says int(0..10)

PR #1271, opened 2023-09-16
Added unit tests and end2end tests for list permitted partitions

anilak11:
Details:

* Added some more cases in unit tests to check filter_args funtionality for list permitted partitions
* Added end2end test case to check the list permitted partitions functionality

### DIFF
--- a/tests/end2end/test_partition.py
+++ b/tests/end2end/test_partition.py
@@ -296,3 +296,96 @@ def test_part_list_os_messages(dpm_mode_cpcs):  # noqa: F811
         assert len(messages) == end - begin + 1
         for message in messages:
             assert message in all_messages
+
+
+# Full set of properties that are common on all types of partitions:
+COMMON_PROPS_LIST = ['name', 'object-uri', 'type', 'status',
+                     'has-unacceptable-status']
+FULL_PROPS_LIST = COMMON_PROPS_LIST + [
+    'initial-memory', 'maximum-memory', 'partition-id', 'processor-mode',
+    'autogenerate-partition-id', 'boot-device', 'boot-network-device',
+    'description', 'hba-uris', 'ifl-processors',
+    'initial-ifl-processing-weight', 'nic-uris', 'partition-link-uris',
+    'storage-group-uris', 'tape-link-uris', 'virtual-function-uris'
+]
+LIST_PERMITTED_PARTITION_TESTCASES = [
+    # The list items are tuples with the following items:
+    # - desc (string): description of the testcase.
+    # - input_kwargs (dict): Input parameters for the function.
+    # - exp_props (List) : Expected properties in the output
+    (
+        "Default parameters",
+        dict(),
+        COMMON_PROPS_LIST,
+    ),
+    (
+        "full_properties",
+        dict(
+            full_properties=True,
+        ),
+        FULL_PROPS_LIST
+    ),
+    (
+        "full_properties and filtering",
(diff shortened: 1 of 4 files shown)


tests/end2end/test_partition.py
```diff
@@ -22,18 +22,19 @@
...
 # pylint: disable=line-too-long,unused-import
-from zhmcclient.testutils import hmc_definition, hmc_session  # noqa: F401, E501
 from zhmcclient.testutils import dpm_mode_cpcs  # noqa: F401, E501
-# pylint: enable=line-too-long,unused-import
-
+from zhmcclient.testutils import hmc_definition, hmc_session  # noqa: F401, E501
 from .utils import skip_warn, pick_test_resources, TEST_PREFIX, \
     standard_partition_props, runtest_find_list, runtest_get_properties
 
+# pylint: enable=line-too-long,unused-import
```
andy-maier, line 36: Why did you move the pylint line? It puts the import from `.utils` into the pylint section where unused import is disabled. That is only necessary for imports that provide pylint fixtures or their dependent fixtures (such as dpm_mode_cpcs, hmc_definition or hmc_session).

If the import from `.utils` has unused imports, they should be removed instead.

tests/end2end/test_partition.py
```diff
@@ -204,3 +205,116 @@ def test_part_crud(dpm_mode_cpcs):  # noqa: F811
...
+        console = client.consoles.console
+
+        part_list = cpc.partitions.list()
+        if not part_list:
+            skip_warn("No partitions on CPC {c} managed by HMC {h} for "
+                      "the user {u}".format(c=cpc.name, h=hd.host,
+                                            u=session.userid))
+        part_list = pick_test_resources(part_list)
+
+        user_name = role_name = TEST_PREFIX + '_test_list_permitted_partitions'
```
andy-maier, line 234: This is a test about listing partitions. There is no need to create or delete any users or password rules.

tests/unit/zhmcclient/test_partition.py
```diff
@@ -18,14 +18,14 @@
 
 from __future__ import absolute_import, print_function
 
-import re
 import copy
+import re
+
 import pytest
+from tests.common.utils import assert_resources
```
andy-maier, line 25: The order of imports so far was grouped into these groups, in order:
* built-in modules
* installed modules
* zhmcclient
* tests

There is no particular technical reason for that; it is just ordered from global to local.

`tests` is a bit special because it is just a local directory but occupies that name in the namespace of packages.

I prefer if we can keep this grouping.

andy-maier (changes requested):
See comments

andy-maier:
@anilak11 Please rebase the PR on the latest master to resolve the conflicts.

tests/end2end/test_partition.py
anilak11: okay, reverted it back

tests/end2end/test_partition.py
anilak11: As discussed, I thought of testing hmc functionality, Now I'm testing the list_permitted_partitions functionality

tests/unit/zhmcclient/test_partition.py
anilak11: I just did a formatting of file on my pycharm IDE, reverted it back and will follow the above instructions from now on

andy-maier (approved):

PR #1283, opened 2023-09-26
Added Console.list_permitted_adapters()

andy-maier:


andy-maier:
I had to do an update that:
* Activated the test for the "firmware-update-pending" property, along with issuing a warning if it does not come back at all on z16 CPCs.
* Fixed an issue in the actual method implementation for the full_properties case.

vkathir82 (approved):

PR #1291, opened 2023-10-06
Added mocked unit test cases for resource property methods

andy-maier:


tests/unit/zhmcclient/test_resource.py
```diff
@@ -642,3 +643,778 @@ def test_get_properties_local(
...
+            'status': 'active',
+            'dpm-enabled': True,
+            'is-ensemble-member': False,
+            'iml-mode': 'dpm',
+            'machine-type': '2964',  # z13
+            'machine-model': 'N10',
+        })
+        self.cpc = self.client.cpcs.find(name='fake-cpc1-name')
+
+    def add_standard_hipersocket(self):
```
EdGue42, line 685: Out of curiosity, why are you setting up a hipersocket adapter?

tests/unit/zhmcclient/test_resource.py
```diff
@@ -642,3 +643,778 @@ def test_get_properties_local(
...
+        # Each list item is a tuple defining a testcase in the following format:
+        # - input_kwargs: keyword input parameters for the function
+        # - delete: Indicates whether to delete the resource before test
+        # - supports_properties: Resource supports 'properties' query parm
+        # - exp_propnames: Expected property names in resource after test
+        # - exp_full_props: Expected value for full_properties attribute
+        # - exp_exc_type: Expected type of exception, or None for success
+
+        (
+            dict(properties=None),
```
EdGue42, line 792: would it make sense to add a comment to each testcase to explain what it is testing?

EdGue42 (approved):

tests/unit/zhmcclient/test_resource.py
andy-maier: Because that is an example of a creatable resource type that does not support the "properties" query parameter. See the docstring of the method.

tests/unit/zhmcclient/test_resource.py
andy-maier: We often have a "description" item in such testcases, but this time I skipped that.
You are right in that we should have it. I'll add it.

andy-maier: Since PR #1292 already merged the content of this PR, the change needs to be done in a future PR. I have created issue TBD for that.

PR #1302, opened 2023-10-25
End2end tests: Logging support; LPAR related support

andy-maier:
Needed for an upcoming PR that adds end2end testcases for LPARs.

Major review points:
* Description of status value meanings in `pull_lpar_status()`: https://github.com/zhmcclient/python-zhmcclient/pull/1302/files#diff-cb055a90f6c2a554d629e2c22aaec5309385084fee4cfa346b067c228f004395R804

### DIFF
--- a/tests/end2end/utils.py
+++ b/tests/end2end/utils.py
@@ -23,13 +23,89 @@ import re
 import random
 import time
 import warnings
+import logging
 import pytest
-from zhmcclient import BaseManager
 
 import zhmcclient
 
 # Prefix used for names of resources that are created during tests
 TEST_PREFIX = 'zhmcclient_tests_end2end'
+
+
+def setup_logging(enable_logging, testfunc_name, log_file=None):
+    """
+    Set up logging for end2end testcases.
+
+    If enable_logging is True, two loggers are created and enabled for logging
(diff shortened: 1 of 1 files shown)


tests/end2end/utils.py
```diff
@@ -664,3 +740,89 @@ def cleanup_and_import_example_certificate(cpc):
...
+    it.
+
+    LPAR status values and their meaning:
+
+    Status         Resources allocated  OS running  Notes
+    ---------------------------------------------------------------------------
+    not-activated  no                   no
+    not-operating  yes                  no
+    operating      yes                  yes         No degradations
+    acceptable     yes                  yes         Some channel degradations
```
andy-maier, line 816: There is still internal discussion going on about the value "acceptable".

anilak11 (approved):
Looks good to me

tests/end2end/utils.py
andy-maier: It seems that a value of "acceptable" is not shown at the WS-API. It is only shown in the SE GUI. I have updated the description accordingly. After this is finally clarified, we can move the description of status values into the user documentation, but for this internal test function this is good enough.

PR #1310, opened 2023-10-30
Updated design description of LPAR activation/load for new 2.16 features

andy-maier:
A direct link to the updated document is: https://github.com/zhmcclient/python-zhmcclient/blob/andy/update-activate-load/design/lpar-status.md

### DIFF
--- a/design/lpar-status.md
+++ b/design/lpar-status.md
@@ -10,13 +10,15 @@ mode; DPM mode is not described here.
 All behaviors and properties are described from a perspective of the WS API
 of the HMC.
 
-Status of this document: Draft, with review comments applied as of 2/22.
+Last updated: 2023-10-30
 
 
 ## Behavior of LPAR activation and loading
 
-This section describes the actual behavior of the "Activate LPAR" and
-"Load LPAR" WS-API operations.
+This section describes the actual behavior of the "Activate Logical Partition"
+and "Load Logical Partition" WS-API operations.
+
+Note: The "Load" WS-API operation introduced with HMC 2.16 is not covered yet.
 
 ### Requirements and facts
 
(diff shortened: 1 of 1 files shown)


design/lpar-status.md
```diff
@@ -31,70 +33,113 @@ This section describes the actual behavior of the "Activate LPAR" and
...
-Notes:
-
 * If a load is performed, the resulting LPAR status will go to 'operating' if
   the loading has succeeded and the control program got started. If the load
   parameters do not specify a valid control program, or if the load parameters
   are not valid for the control program, the LPAR status will go to
   'not-operating'.
 
-### Behavior of "Load LPAR" operation
+### Behavior of "Load Logical Partition" operation
```
Emily1799, line 142: This header is misleading, because you're still talking about what is "under the hood" the 'Activate" operation. Load Logical Partition in the API is a separate operation that refers to an ECKD-CCW load

design/lpar-status.md
```diff
@@ -212,73 +259,41 @@ book.
...
-| Activate w/ force (load pr. modified)       | operating        | operating (react+reload)             | No          |
+   - What is the "system-defined Logical Partition group"?
 
-### SSC-type LPAR
+2. 'load-type' parameter of activate operation for SSC/zAware partitions
 
-Same as Linux-type: Yes?
+   The new 'load-type' parameter of the "Activate Logical Partition" operation
+   in HMC>=2.16 is silent about whether it can be used for LPARs whose
+   image activation profile specifies 'activation-mode' as 'ssc' or 'zaware'.
```
Emily1799, line 293: It cannot be used if the image's profile's activation-mode is ssc or zaware; there's a check to prevent it synchronously. It returns a bad request reason code 306

design/lpar-status.md
andy-maier: Actually this section does want to talk about the "Load Logical Partition" operation. The implicit load done as part if the "Activate Logical Partition" operation is covered n the same-named chapter.

design/lpar-status.md
andy-maier: Thanks for the clarification. I'll update the document.

design/lpar-status.md
Emily1799: OK. Then it seems a bit confusing to name just one of the load operations then? Unless you're referring to the naming of one of your modules, in which case, sorry for my confusion.

andy-maier: The load operations are meant to be the HMC operations. At this point, the document only describes the "Load Logical Partition" HMC operation, because that is the one the zhmc_lpar Ansible module uses to perform a load, so it is of greater interest than the others. I am aware of the newer "Load" HMC operation and at some point the zhmc_lpar Ansible module will migrate to use that one.

Emily1799: If you're only using Load Logical Partition, does that mean that you're not presently handling SCSI or NVMe loads right now? That would be SCSI Load / SCSI Dump / NVMe Load / NVMe Dump.

Thrilled to hear you'll be migrating to the new Load (`POST /api/logical-partitions/{logical-partition-id}/operations/load-program`)  operation

andy-maier: There are zhmcclient methods for SCSI and NVMe load and dump. Just this particular document does not cover them.

The only load operation not yet supported is the newer "load-program".

PR #1343, opened 2023-11-17
Added support for retrievel of firmware from an FTP server

andy-maier:


schaufle (approved):
Looks good to me. Thanks for the very quick turnaround!

andy-maier:
Update needed:
* Since the bundle-level is now optional, the method descriptions need to be updated to explain that.

PR #1345, opened 2023-11-22
Add load-from-ftp operation

smitterl:
Add load-from-ftp operation.

The session-credential can be used to connect to the API message broker. Store it as a new property.

Add script to load-from-ftp with a config file.


smitterl:
Seems load-from-ftp has already been added so closing this.

PR #1346, opened 2023-11-22
Add example for load-from-ftp

smitterl:
Add example to use load-from-ftp for automated LPAR installation.

The session-credential can be used to connect to the API message broker. Store it as a new property.

Tested on a z16 with API 2.26

Suggested-by: Vance Morris <vmorris@redhat.com>
Signed-off-by: Sebastian Mitterle <smitterl@redhat.com>

### DIFF
--- a/examples/install_from_ftp.py
+++ b/examples/install_from_ftp.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright Red Hat
+#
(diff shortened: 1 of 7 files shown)


andy-maier (changes requested):
Sebastian, thanks for this contribution.

The test workflow fails because the zhmcclient mock support does not return the session-credentials property from its mocked logon operation. Can you please add that as well?

andy-maier:
I may be asking for too much here. I'll create issues for the missing mock support, and the use of session-credentials and then will integrate your example script. Stay tuned.

smitterl:
> I may be asking for too much here. I'll create issues for the missing mock support, and the use of session-credentials and then will integrate your example script. Stay tuned.

Hi @andy-maier. Sorry for the late answer. I just added the commit.

andy-maier (approved):

PR #1362, opened 2023-11-29
Added support for SE firmware upgrade related operations

andy-maier:
The PR adds support for these HMC operations:
* "CPC Install and Activate"
* "CPC Delete Retrieved Internal Code"

Particular review points:
* The way the operation parameters are represented in Python.
* Returning the job message in `Cpc.install_and_activate()`
* Statement about not downloading updates from the IBM support site in `Cpc.install_and_activate()`
* Description of which updates are deleted when `ec_levels` is specified in `Cpc.delete_retrieved_internal_code()`

### DIFF
--- a/zhmcclient/_cpc.py
+++ b/zhmcclient/_cpc.py
@@ -2201,6 +2201,262 @@ class Cpc(BaseResource):
         return result
 
     @logged_api_call
+    def install_and_activate(
+            self, bundle_level=None, ec_levels=None,
+            install_disruptive=False, wait_for_completion=True,
+            operation_timeout=None):
+        """
+        Installs and activates firmware on the Support Element (SE) of this CPC.
+
+        Brief summary of firmware levels:
+
+            The firmware is segmented into different subsystems called
+            "Engineering Change" (EC), or sometimes "EC stream". An EC stream
+            is identified by an EC number (e.g. "P30719").
+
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: c36bdd7703ce14862aa65b380676ff17b2668359

---

zhmcclient/_cpc.py
```diff
@@ -2200,6 +2200,262 @@ def single_step_install(
...
+
+        * Object-access permission to this CPC.
+        * Task permission to the "Change Internal Code" task.
+
+        Parameters:
+
+          bundle_level (string): Name of the bundle to be installed on the SE
+            of this CPC (e.g. 'S51').
+
+            This parameter is mutually exclusive with `ec_levels`. If both
```
schaufle, line 2272: The code below does not check the mutual exclusive requirement. Is it the design point of zhmcclient to let this pass and let the underlying WSAPI implementation detect this?

schaufle (commented):
Looks good to me. The one comment I made is probably "works as designed"

zhmcclient/_cpc.py
andy-maier: Yes, that is done for most operations in the zhmcclient. The idea is not to double check because this would shorten only the error path, but make the good path slightly (very minimally though) longer.

PR #1364, opened 2023-12-13
Tolerate restarts while waiting for job completion

abhirams-1:
Changes:
* Some jobs like single-step-install require restarting the HMC several times. Requests for job status will return ConnectionError while HMC is restarting. This will be ignored to tolerate HMC restarts that will happen while waiting for completion of the job.

This solves #1365 

Signed-off-by: Abhiram Santhosh <abhiram.santhosh1@ibm.com>

### DIFF
--- a/zhmcclient/_session.py
+++ b/zhmcclient/_session.py
@@ -1773,7 +1773,6 @@ class Job(object):
           :exc:`~zhmcclient.ParseError`
           :exc:`~zhmcclient.ClientAuthError`
           :exc:`~zhmcclient.ServerAuthError`
-          :exc:`~zhmcclient.ConnectionError`
           :exc:`~zhmcclient.OperationTimeout`: The timeout expired while
             waiting for job completion.
         """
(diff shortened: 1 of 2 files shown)


zhmcclient/_session.py
```diff
@@ -1785,7 +1784,13 @@ def wait_for_completion(self, operation_timeout=None):
             start_time = time.time()
 
         while True:
-            job_status, op_result_obj = self.check_for_completion()
+            try:
+                job_status, op_result_obj = self.check_for_completion()
+            except ConnectionError:
+                HMC_LOGGER.debug("Got ConnectionError while waiting for "
+                                 "completion of job %s. This could be "
+                                 "because HMC is restarting.", self.uri)
```
andy-maier, line 1792: I suggest that the log message makes it clear that a retry is performed. Something like: "Retrying after ConnectionError while waiting for ...."

zhmcclient/_session.py
```diff
@@ -1773,7 +1773,6 @@ def wait_for_completion(self, operation_timeout=None):
           :exc:`~zhmcclient.ParseError`
           :exc:`~zhmcclient.ClientAuthError`
           :exc:`~zhmcclient.ServerAuthError`
-          :exc:`~zhmcclient.ConnectionError`
```
andy-maier, line 1776: Please add a change log entry for this change. The change log entry should make it clear that Session.wait_for_completion() now retries in case of connection errors and no longer raises ConnectionError.

andy-maier (changes requested):
See comments

zhmcclient/_session.py
abhirams-1: Done

zhmcclient/_session.py
abhirams-1: Done

andy-maier (approved):

PR #1452, opened 2024-02-19
Added Nic.backing_port() method

andy-maier:

### DIFF
--- a/tests/end2end/test_nic.py
+++ b/tests/end2end/test_nic.py
@@ -22,6 +22,7 @@ and delete test partitions with NICs.
 from __future__ import absolute_import, print_function
 
 import warnings
+import random
 import pytest
 from requests.packages import urllib3
 
@@ -32,7 +33,12 @@ from zhmcclient.testutils import dpm_mode_cpcs  # noqa: F401, E501
 # pylint: enable=line-too-long,unused-import
 
 from .utils import skip_warn, pick_test_resources, TEST_PREFIX, \
-    standard_partition_props, runtest_find_list, runtest_get_properties
+    standard_partition_props, runtest_find_list, runtest_get_properties, \
+    setup_logging
+
+# Logging for zhmcclient HMC interactions and test functions
+LOGGING = False
+LOG_FILE = 'test_nic.log'
 
 urllib3.disable_warnings()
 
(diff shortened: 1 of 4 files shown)


tests/end2end/test_nic.py
```diff
@@ -241,3 +247,145 @@ def test_nic_crud(dpm_mode_cpcs):  # noqa: F811
...
+    Test Nic.backing_port() for port-based NICs (e.g. RoCE, CNA).
+    """
+    if not dpm_mode_cpcs:
+        pytest.skip("HMC definition does not include any CPCs in DPM mode")
+
+    logger = setup_logging(LOGGING, 'test_nic_backing_port_port_based',
+                           LOG_FILE)
+    logger.debug("Entered test function")
+
+    cpc = random.choice(dpm_mode_cpcs)
```
EdGue42, line 264: Hrmrm. On the one hand, I appreciate the idea to limit the execution to a single CPC, and to then pick that one randomly. 
But: random adds an element of random. So, worst case, your test passes / fails randomly (because of issues with the underlying CPCs). On the other hand, the underlying functionality should be stable, so "which CPC was picked" shouldn't have such effects.

tests/end2end/test_nic.py
```diff
@@ -241,3 +247,145 @@ def test_nic_crud(dpm_mode_cpcs):  # noqa: F811
...
+    logger.debug("Testing with CPC %s", cpc.name)
+    client = cpc.manager.client
+
+    port_nics = []
+    partitions = cpc.partitions.list()
+    for partition in random.sample(partitions, k=len(partitions)):
```
EdGue42, line 270: I don't get why you iterate over all partitions, but random order? Isn't it "random" anyway already?

tests/end2end/test_nic.py
andy-maier: It is always possible that on machines used for FW development/test, something does not work. On stable machines (like our A224/T224/M224M in POK) it should work reliably. In the testcase, it is not practical to have knowledge about specific CPCs. In case something does not work, a user running end2end tests would need to use a different HMC. So I would say unstable development CPCs / HMCs is something we can deal with.

tests/end2end/test_nic.py
andy-maier: I want to avoid iterating through all NICs of all partitions, just to pick one for the test. So I am limiting the iteration of NICs until we have 5 (or less if the loop exhausts before that). In order not to use the same partitions over and over again, I am randomizing the order in which the partitions are iterated, so we get the 5 NICs from potentially any partition.

PR #1457, opened 2024-03-14
Support additional-properties in list permitted partitions

simon-spinner:
Latest HMC versions support to include additional properties in the results. This will allow us to optimize list partitions in zhmccli.


andy-maier (approved):

andy-maier:
I ran an end2end test on:
* TESTHMC=A224 (2.16 at API version 4.10) - new test was executed and passed
* TESTHMC=T01 (2.15) - new test was skipped (as it should be)
I will merge the PR.

PR #1474, opened 2024-05-10
Used Queue for STOMP listener handover; Improved connection loss recovery

andy-maier:

This change was tested with:
* The example script "show_os_messages.py"
* The zhmc-log-forwarder (as of its PR https://github.com/zhmcclient/zhmc-log-forwarder/pull/14)

In both cases, the default STOMP retry/timeout config was used (i.e. with STOMP heartbeating enabled), the network was disconnected and connected again, and the notification receiver detected the disconnect properly and also reconnected properly.

### DIFF
--- a/changes/1502.feature.rst
+++ b/changes/1502.feature.rst
@@ -1,0 +1,40 @@
+Improved the notification support in several ways:
+
+- Replaced the event-based handover of a single item from the notification
+  listener thread to the caller's thread with a Python Queue, for better
+  reliability. It turned out that messages could have been lost in some cases
+  with the previous design.
+
+- The 'NotificationReceiver.notifications()' method now continues running
+  when there are no notifications, and only ever returns when
+  'NotificationReceiver.close()' is called (by some other thread).
+
+- Added methods 'connect()' and 'is_connected()' to the 'NotificationReceiver'
+  class. The init method of 'NotificationReceiver' no longer connects,
+  but the 'notifications()' method now calls 'connect()', so overall this is
+  compatible with the prior behavior.
+
+- Added new exceptions 'NotificationConnectionError' and
+  'NotificationSubscriptionError' that may be raised by some
+  'NotificationReceiver' methods.
+
+- Documented the stomp-py exceptions that can be raised from
+  'NotificationReceiver' methods.
+
+- Added proper detection of STOMP connection loss if STOMP heartbeating is
+  enabled. The connection loss is surfaced by raising
+  'NotificationConnectionError' in 'NotificationReceiver.notifications()'.
+  This allows users to retry 'NotificationReceiver.notifications()' upon
+  connection loss.
+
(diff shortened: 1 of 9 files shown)


andy-maier:
Rebased on master after merging PR #1510 .

changes/1502.incompatible.rst
```diff
@@ -0,0 +1,6 @@
+The 'NotificationReceiver.notifications()' method now continues running when
+there are no notifications, and only ever returns when exceptions are raised.
```
andy-maier, line 2: Change to:
```
only ever returns when 'NotificationReceiver.close()'  is called.
```

changes/1502.incompatible.rst
```diff
@@ -0,0 +1,6 @@
...
+Before this change, the method returned when there were no notfications, so it
+had to be invoked by the user in a loop. While such user code would not break
+with this change, it should still be adjusted to accomodate that the method is
+now not expected to ever return.
```
andy-maier, line 6: There is a second incompatibility: Previously, we did not document that `stomp.exception.StompException` can be raised from several methods.

changes/1502.feature.rst
```diff
@@ -0,0 +1,36 @@
...
+  increases from this minimum with as the heartbeat receive time gets larger.
+
+- Added the missing event methods to the internal '_NotificationListener'
+  class in case they are ever invoked (needed due to lazy importing of
+  stomp-py).
+
+- Added more log messages around STOMP connect / disconnect.
+
+- Documented the stomp-py exceptions that can be raised from public
+  zhmcclient methods.
```
andy-maier, line 36: Add "Issue:" in a separate line so the issue link gets added there and not to the last item in the list.

tests/unit/zhmcclient/test_notification.py
```diff
@@ -126,21 +146,38 @@ def mock_add_message(self, headers, message):
 
     def mock_start(self):
         """Start the STOMP message sender thread."""
-        assert self._state_connected
+        # assert self._state_connected
```
EdGue42, line 149: clean code: what is the purpose of commenting out code here (and below)?
If this is a temporary thing, consider to add a TODO / FIXME explanation.

If this is meant to stay in, add a reason why it is commented but needs to stay in the file.

EdGue42 (approved):

tests/unit/zhmcclient/test_notification.py
andy-maier: Thanks for catching this. This was in fact a try-out that I forgot to resolve in one or the other way. I think the mock_start/end functions should not be dependent on the connection state, and have removed the commented-out assertions.

changes/1502.incompatible.rst
andy-maier: Added

changes/1502.feature.rst
andy-maier: done

andy-maier:
Resolved all review comments.

PR #1482, opened 2024-05-19
Test of check for no change fragment file

andy-maier:


andy-maier:
This test does not work since the target branch of the PR is used to compare with. Tested it locally, though.

PR #1486, opened 2024-05-20
Migrated to pyproject.toml

andy-maier:


Makefile
```diff
@@ -535,9 +542,10 @@ ifeq ($(PLATFORM),Windows_native)
 	@echo "Makefile: Warning: Skipping the checking of missing dependencies of site-packages directory on native Windows" >&2
 else
 	@echo "Makefile: Checking missing dependencies of some development packages in our minimum versions"
-	@rc=0; for pkg in $(check_reqs_packages); do dir=$$($(PYTHON_CMD) -c "import $${pkg} as m,os; dm=os.path.dirname(m.__file__); d=dm if not dm.endswith('site-packages') else m.__file__; print(d)"); cmd="pip-missing-reqs $${dir} --requirements-file=minimum-constraints.txt"; echo $${cmd}; $${cmd}; rc=$$(expr $${rc} + $${?}); done; exit $${rc}
+	cat minimum-constraints-develop.txt minimum-constraints-install.txt >tmp_minimum-constraints.txt
```
EdGue42, line 545: minor: add space after ">" for readability

EdGue42 (approved):

Makefile
andy-maier: I never do that. Is that a common rule?

If we change it, it should be done consistently. We have 20+ redirections in the Makefile, none of which is followed by a blank.

EdGue42: Never mind, I guess that is purely personal style then.

PR #1481, opened 2024-05-19
Migrated changelog to use towncrier

andy-maier:
For details, see the commit message.

This is a first step for simplifying the procedures for releasing a version and starting a new version. For the complete proposal, see issue #1485

PR #1487, opened 2024-06-03
Remove password property before updating local user object

abhirams-1:
Changes:
- User object(python object, not HMC) does not need to store password. Filter and remove 'password' from the propertied before updating the local object during user create and update.

Fixes https://github.com/zhmcclient/zhmc-ansible-modules/issues/967

### DIFF
--- a/zhmcclient/_user.py
+++ b/zhmcclient/_user.py
@@ -178,6 +178,7 @@ class UserManager(BaseManager):
         # returned props should overwrite the input props:
         props = copy.deepcopy(properties)
         props.update(result)
+        props.pop('password', None)
         name = props.get(self._name_prop, None)
         uri = props[self._uri_prop]
         user = User(self, uri, name, props)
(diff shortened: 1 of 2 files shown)


zhmcclient/_user.py
```diff
@@ -275,6 +277,8 @@ def update_properties(self, properties):
         # HTTPError to be raised in the POST above, so we assert that here,
         # because we omit the extra code for handling name updates:
         assert self.manager._name_prop not in properties
+        if properties is not None:
+            properties.pop('password', None)
```
andy-maier, line 281: I don't think the removal of the "password" property should be done from the `properties` input parameter, because of the side effect of changing the input parameter.

I suggest to remove the "password" property from the copy of the `properties` input parameter which is created one line down.

Also, there should be an entry in the change log about this change. It is an incompatible change since up to now, users were able to store the password in the User object and now that no longer works.

andy-maier (changes requested):

zhmcclient/_user.py
abhirams-1: Done

zhmcclient/_user.py
```diff
@@ -211,6 +211,8 @@ def __init__(self, manager, uri, name=None, properties=None):
         assert isinstance(manager, UserManager), \
             "Console init: Expected manager type {}, got {}" \
             .format(UserManager, type(manager))
+        if properties is not None:
+            properties.pop('password', None)
```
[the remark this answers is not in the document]
abhirams-1: Done

andy-maier (approved):

andy-maier:
I updated the PR by transforming the change log entry into a change fragments file, since PR #1481 was merged meanwhile.

PR #1523, opened 2024-06-14
Dropped support for Python < 3.8

andy-maier:
The different activities to drop Python < 3.8 have been kept separate in this PR.


EdGue42 (approved):
Ignored all ".py" files; assuming that they were changed automatically by tooling.
Rest looks good to me.

PR #1529, opened 2024-06-16
Added support for running the 'bandit' checker

andy-maier:


Makefile
```diff
@@ -515,6 +520,13 @@ $(done_dir)/safety_install_$(pymn)_$(PACKAGE_LEVEL).done: $(done_dir)/develop_$(
 	echo "done" >$@
 	@echo "Makefile: Done running Safety for install packages"
 
+$(done_dir)/bandit_$(pymn)_$(PACKAGE_LEVEL).done: $(done_dir)/develop_$(pymn)_$(PACKAGE_LEVEL).done Makefile $(check_py_files)
+	@echo "Makefile: Running Bandit"
+	-$(call RM_FUNC,$@)
+	bandit -s B101 -r -l $(package_name) $(mock_package_name)
```
EdGue42, line 526: minor: add a comment what "-s B101" is doing?

zhmcclient/_utils.py
```diff
@@ -73,7 +73,7 @@
 # For CPCs in any  mode and resources independent of CPCs:
 RC_CONSOLE = 'console'
 RC_CPC = 'cpc'
-RC_PASSWORD_RULE = 'password-rule'
+RC_PASSWORD_RULE = 'password-rule'  # nosec B105
```
EdGue42, line 76: I assume B105 here ... is something else than the B101 that you used in the Makefile above?

EdGue42 (approved):

Makefile
andy-maier: I cannot explain every option of every command with a comment.
Maybe we should start using a config file for bandit (in which it is much easier to have explanatory comments).

zhmcclient/_utils.py
andy-maier: Yes. "nosec" is for bandit, "pylint" for pylint, and "noqa" for flake8.

Makefile
EdGue42: If you think that there will be many changes to these options, that might be a good solution.
But if this is rather static (and unlikely to be changed frequently), I would keep it as is.

andy-maier: There is always the possibility of changes, and commenting it would indeed be good. I have added a bandit config file.

EdGue42 (approved):

PR #1531, opened 2024-06-16
Added support for running the 'ruff' checker

andy-maier:


EdGue42 (approved):
Should there be some guidance in a help text explaining when to use "ruff" versus "lint"? Having two linters without explanations might lead to "why 2" questions?

andy-maier:
We do not use "lint". We already use flake8 and pylint, and now add ruff. ruff plans to replace flake8, but they are not yet there.

PR #1537, opened 2024-06-18
Adjusted publish workflow to new tags-ignore construct

andy-maier:
No review needed.


andy-maier:
It turns out that `tags-ignore` and `tags` cannot be used together, so the old way of doing it was supposedly correct (see https://github.com/orgs/community/discussions/25114), but it still triggered the workflow when pushing a release start tag.

PR #1539, opened 2024-06-18
 Test: Fixes and improvements in end2end_mocked tests and zhmcclient mock support

andy-maier:

### DIFF
--- a/tests/end2end/mocked_inventory.yaml
+++ b/tests/end2end/mocked_inventory.yaml
@@ -1,0 +1,49 @@
+---
+# HMC inventory file with mocked HMCs for end2end tests run by GitHub Actions
+#
+# This file defines only mocked HMCs (using the zhmcclient mock
+# support) for the purpose of end2end tests against these HMCs.
+# The file must have the format defined in
+# zhmcclient/testutils/hmc_inventory_file.py of the python-zhmcclient project.
+#
+# HMC inventory files conform to the format of HMC inventory files in YAML
+# format and define specific additional variables for HMCs.
+#
+# Brief description of the HMC inventory file format:
+#
+#   all:  # the top-level HMC group
+#     hosts:
+#       <hmc_name>:  # DNS hostname, IP address, or nickname of HMC
+#         description: <string>
+#         contact: <string>
+#         access_via: <string>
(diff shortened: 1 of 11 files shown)


tests/end2end/mocked_inventory.yaml
```diff
@@ -0,0 +1,49 @@
...
+#         vars: ...
+#         children: ...
+
+all:
+  hosts:
+    mocked_hmc_z16:
+      description: "Mocked HMC with z16 in classic mode and z16 in DPM mode"
+      mock_file: "mocked_hmc_z16.yaml"
+      cpcs:
+        CPC1:
```
EdGue42, line 40: minor: why not name them "CLASSIC" and "DPM" instead just 1 and 2 ?

tests/end2end/test_storage_volume.py
```diff
@@ -203,6 +203,9 @@ def test_stovol_crud(dpm_mode_cpcs):  # noqa: F811
             stovol = stogrp.storage_volumes.create(stovol_input_props)
 
             for pn, exp_value in stovol_input_props.items():
+                assert pn in stovol.properties, (
+                    f"Input property {pn!r} is not in created storage volume:\n"
+                    f"{stovol!r}")
```
EdGue42, line 208: consistency: you are using the new f"{}" style here, but keep the old style in other places? 
Maybe run that "update" script on the files you modify?

EdGue42 (approved):

tests/end2end/mocked_inventory.yaml
andy-maier: Good idea. Done (Used CPC_DPM and CPC_CLA)

tests/end2end/test_storage_volume.py
andy-maier: For reasons I do not understand yet, the pyupgrade tool I used for the upgrade of the code to Python 3.8 converted only a subset of the format strings to f-strings. I would like to convert the remaining ones in a separate PR.

andy-maier:
Updated to reflect the review comments.

EdGue42 (approved):

PR #1576, opened 2024-06-27
Test: Transitioned to label 'no change log' for PRs without need for change log

andy-maier:

Particular review points:
* With this approach, the label needs to be set while creating the PR. If that is forgotten, the check fails and then after adding the label, a dummy push is needed to trigger the test workflow again. The question is, is that too cumbersome in the handling.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,9 +112,6 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - name: Make sure this change adds or changes at least one change fragment file
-      run: |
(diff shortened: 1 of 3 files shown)


EdGue42 (approved):
Looks good technically, but I share the doubts about the complex workflow. 
I regularly forget to labels upon creation.

andy-maier:
Then let's discuss in our next meeting.

simon-spinner (approved):

andy-maier:
We discussed the check for fragment files, and decided to remove the entire check, because it creates too much overhead. Missing change log entries then will need to be discovered during review, as it was before we introduced change fragment files.

PR #1578, opened 2024-06-27
Test changelog checking - no change file and no "no change log" label

andy-maier:
Note: This PR needs to be rebased to master once its underlying PR #1576 is merged to actually perform the real test.


andy-maier:
No longer needed, since we removed the change fragment check from the test workflow.

PR #1663, opened 2024-09-15
Debug coveralls issue

andy-maier:


andy-maier:
I created https://github.com/actions/setup-python/issues/942 for the issue.

Solution will be to no longer specify a login shell for invoking coveralls.

PR #1681, opened 2024-09-27
Test coverage with login shell

andy-maier:
Not for merging, but for testing whether the issue is now resolved on GitHub. See https://github.com/actions/setup-python/issues/942


andy-maier:
Using the login shell works again. Apparently that was an issue within GitHub Actions.

PR #1685, opened 2024-10-05
Support for Partition Links

andy-maier:

The PR is ready to be merged.

### DIFF
--- a/tests/end2end/utils.py
+++ b/tests/end2end/utils.py
@@ -23,6 +23,9 @@ import random
 import time
 import warnings
 import logging
+from copy import deepcopy
+import pprint
+from collections.abc import Mapping, MappingView
 import pytest
 
 import zhmcclient
@@ -113,6 +116,36 @@ class End2endTestWarning(UserWarning):
     Python warning indicating an issue with an end2end test.
     """
     pass
+
+
+def assert_properties(act_obj, exp_obj):
+    """
+    Check that actual properties match expected properties.
+
+    The expected properties may specify only a subset of the actual properties.
+    Only the expected properties are checked.
+
+    The property values may have any type, including nested dictionaries and
+    lists. For nested dictionaries and lists, each item is matched recursively.
+
+    Parameters:
+      act_obj (dict): The actual object. Initially, a dict with properties.
+      exp_obj (dict): The expected object. Initially, a dict with properties.
+    """
+    if isinstance(exp_obj, dict):
+        for name, exp_value in exp_obj.items():
+            assert name in act_obj, (
+                f"Expected property {name!r} is not in actual properties:\n"
+                f"{act_obj!r}")
+            act_value = act_obj[name]
+            assert_properties(act_value, exp_value)
+    elif isinstance(exp_obj, list):
+        for i, exp_value in enumerate(exp_obj):
+            act_value = act_obj[i]
+            assert_properties(act_value, exp_value)
+    else:
+        assert act_obj == exp_obj, (
+            f"Unexpected value: {act_obj!r}; Expected: {exp_obj!r}")
 
 
 def assert_res_props(res, exp_props, ignore_values=None, prop_names=None):
@@ -632,6 +665,31 @@ def _skipif_api_feature_not_on_cpc_and_hmc(feature, cpc):
         skip_warn(f"API feature {feature} not available on HMC {console.name}")
 
 
+def skipif_no_partition_link_feature(cpc):
+    """
+    Skip the test if not all of the Partition Link related API features are
+    enabled for the specified CPC, or if the CPC does not yet support all of
+    them:
+
+    * "dpm-smcd-partition-link-management"
+    * "dpm-hipersockets-partition-link-management"
+    * "dpm-ctc-partition-link-management"
+
+    Note that there are code levels for z16 where only SMC-D is enabled, but not
+    Hipersockets and CTC.
+    """
+    has_all_features = (
+        has_api_feature(
+            "dpm-smcd-partition-link-management", cpc) and  # noqa: W504
+        has_api_feature(
+            "dpm-hipersockets-partition-link-management", cpc) and  # noqa: W504
+        has_api_feature(
+            "dpm-ctc-partition-link-management", cpc))
+    if not has_all_features:
+        skip_warn("The partition link related API features are not all enabled "
+                  f"or not all supported on CPC {cpc.name}")
+
+
 def has_api_feature(feature, cpc):
     """
     Returns True if the given API feature is available on the specified CPC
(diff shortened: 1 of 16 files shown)


zhmcclient/_partition_link.py
```diff
@@ -0,0 +1,752 @@
...
+            waiting for completion of the operation.
+        """
+        body = {}
+        pl_type = self.get_property('type')
+        if pl_type == 'ctc':
+            body['added-partition-uris'] = [partition.uri]
+            body.update(properties)
+            # properties:
+            #   'devices-per-path'
+            #   'added-paths'
```
simon-spinner, line 504: As part of the `attach_to_partition` you should not need to specify `added-paths` and `devices-per-path`. This information needs to be provided once when creating the partition link to represent the physical connectivity. Unless you do changes to the physical connectivity of your system, you don't need to specify them in the Modify PL call again.

It gets complicated if we want to allow to specify device numbers for the newly attached partitions. Then we would have to dig into the `modified-paths`. But for initial support that's probably not necessary as we automatically generate device numbers if not specified.

zhmcclient/_partition_link.py
```diff
@@ -0,0 +1,752 @@
...
+            list_uri, result_prop, full_properties, filter_args, None)
+
+    @logged_api_call
+    def create(self, properties=None):
+        """
+        Create a partition link from the input properties.
+
+        The input properties may specify an initial attachment of the partition
+        link to one or more partitions.
+
```
simon-spinner, line 213: For CTC it is required to specify a `paths` array on initial creation describing the physical connectivity between the Ficon adapters (which adapter is connected to which). We may want to state that here (although of course you can find that information also in the HMC WS-API book.

andy-maier: Added.

zhmcclient/_partition_link.py
andy-maier: I have moved the property description into the docstring, and have applied this comment.

zhmcclient/_partition_link.py
```diff
@@ -0,0 +1,786 @@
...
+                if part_uri not in part_uri_list:
+                    part_uri_list.append(part_uri)
+                    part_list.append(part)
+        return part_list
+
+    def attach_to_partition(
+            self, partition, properties, wait_for_completion=True,
+            operation_timeout=None):
+        """
+        Attach this partition link to a partition, creating corresponding
```
simon-spinner, line 442: We have the attach tape link/storage group operations on the Partition resource. I am wondering whether it makes sense to move the attach partition link operation there as well for consistency.

andy-maier: I like the idea, but it is not clear to me whether that can be done for CTC. Let's discuss.

andy-maier: I have moved them and split them up between network and CTC.

tests/end2end/utils.py
```diff
@@ -632,6 +662,25 @@ def _skipif_api_feature_not_on_cpc_and_hmc(feature, cpc):
         skip_warn(f"API feature {feature} not available on HMC {console.name}")
 
 
+def skipif_no_partition_link_feature(cpc):
+    """
+    Skip the test if the "dpm-smcd-partition-link-management" API feature is
+    not enabled for the specified CPC, or if the CPC does not yet support it.
+
+    Note that the three partition-link related API features are always all
+    enabled or all disabled:
```
simon-spinner, line 671: Not fully correct. There are code levels for z16 where only SMC-D is enabled, but not HiSo and CTC.

zhmcclient/_partition.py
```diff
@@ -1550,6 +1550,519 @@ def list_attached_storage_groups(self, full_properties=False):
...
+        Authorization requirements:
+
+        * Object-access permission to this partition.
+        * Object-access permission to the partition link.
+        * Task permission to the "Partition Link Details" task.
+
+        Parameters:
+
+          partition_link (:class:`~zhmcclient.PartitionLink`): The partition
+            link that will be attached to this partition.
```
simon-spinner, line 1912:
```suggestion
            link that will be detached from this partition.
```

zhmcclient/_partition.py
```diff
@@ -1550,6 +1550,519 @@ def list_attached_storage_groups(self, full_properties=False):
...
+        if type:
+            filter_args['type'] = type
+        if name:
+            filter_args['name'] = name
+        if state:
+            filter_args['state'] = state
+
+        pl_list = console.partition_links.list(
+            filter_args=filter_args,
+            additional_properties=additional_properties)
```
simon-spinner, line 2035: Can we always add 'paths' and 'bus-connections' to additional_properties for improved performance as they are used for the filtering below?

zhmcclient/_partition_link.py
```diff
@@ -0,0 +1,613 @@
...
+
+        HMC/SE version requirements:
+
+        * SE version >= 2.16.0
+        * for technology-specific support, see the API features described in
+          :ref:`Partition Links`
+
+        Authorization requirements:
+
+        * Object-access permission to the CPC that will be associated with
```
simon-spinner, line 547: You don't need object access permissions for the CPC on modify for an existing partition link. That's only the case when you create a new partition link.

tests/end2end/utils.py
andy-maier: Changed the function to check that all PL related features are enabled. 
DONE.

zhmcclient/_partition.py
andy-maier: DONE

zhmcclient/_partition_link.py
andy-maier: DONE

zhmcclient/_partition.py
andy-maier: DONE

zhmcclient/_partition.py
```diff
@@ -1550,6 +1550,529 @@ def list_attached_storage_groups(self, full_properties=False):
...
+                part1.attach_network_link(
+                    hs_link, nic_property_list=[
+                        {
+                            'device-number': '1300',
+                            'vlan-id': '53',
+                        }
+                    ])
+        """
+        body = {}
+        # pl_type = partition_link.get_property('type')
```
simon-spinner, line 1655: Why is this code commented out?

simon-spinner (approved):

zhmcclient/_partition.py
andy-maier: I think this was a remainder from the initial implementation of the attach method where it tried to support both CTC and network links. I have removed it. DONE.

PR #1740, opened 2025-01-22
Support for MFA Server Definitions

andy-maier:


EdGue42 (approved):
Looking good (didn't check the test in depth though)

PR #1742, opened 2025-01-22
Added retries when HMC returns busy (HTTP 409)

andy-maier:

### DIFF
--- a/zhmcclient/_partition_link.py
+++ b/zhmcclient/_partition_link.py
@@ -519,7 +519,8 @@ class PartitionLink(BaseResource):
             uri=f'{self.uri}/operations/delete',
             resource=self,
             body=body,
-            wait_for_completion=True)
+            wait_for_completion=True,
+            busy_retries=1, busy_wait=2)
 
         # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


zhmcclient/_partition_link.py
```diff
@@ -577,7 +578,8 @@ def update_properties(self, properties):
             uri=f'{self.uri}/operations/modify',
             resource=self,
             body=properties,
-            wait_for_completion=True)
+            wait_for_completion=True,
+            busy_retries=1, busy_wait=2)
```
anilak11, line 582: Any reason why these specific values for delete & update partition link ?

anilak11 (approved):
Otherwise looks good to me

zhmcclient/_partition_link.py
andy-maier: I wanted the method to work with its default values, and since I did encounter cases where busy was returned, I thought that 1 retry should be default. The busy waiting time is just a guess, but it worked well enough to make the busy conditions no longer show up. Let me know if you have different suggestions for the default values.

simon-spinner (approved):

PR #1759, opened 2025-02-20
Filter out non-FCP storage groups in test_virtual_storage_resource.

simon-spinner:
The test otherwise fails with a 404.1 if the system has non-FCP storage groups.


andy-maier (approved):

andy-maier:
Rolled back into 1.19 with PR #1787.

PR #1776, opened 2025-03-20
Improved performance of LPAR/adapter lookup during metrics; Fixed exception msg

andy-maier:

I tested the performance improvement using the zhmc prometheus exporter: On the HMC of T28 which manages 7 CPCs all in classic mode with a total of 44 LPARs, the elapsed time for the first data collection (where the resource lookup is performed) went down from 59 sec to 12 sec.

### DIFF
--- a/zhmcclient/_console.py
+++ b/zhmcclient/_console.py
@@ -990,6 +990,8 @@ class Console(BaseResource):
         HMC/SE version requirements:
 
         * HMC version >= 2.16.0
+        * The adapters of CPCs with SE version < 2.16.0 that are not enabled
+          for DPM are not included in the result.
 
         Authorization requirements:
 
--- a/zhmcclient/_metrics.py
+++ b/zhmcclient/_metrics.py
@@ -85,6 +85,10 @@ __all__ = ['MetricsContextManager', 'MetricsContext', 'MetricGroupDefinition',
            'MetricDefinition', 'MetricsResponse', 'MetricGroupValues',
            'MetricObjectValues']
 
+# HMC API version for certain HMC versions
+API_VERSION_HMC_2_14_0 = (2, 20)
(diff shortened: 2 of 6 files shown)


zhmcclient/_metrics.py
```diff
@@ -906,21 +898,13 @@ def resource(self):
...
-            for cpc in cpc_list:
-                adapter_managers.append(cpc.adapters)
-                try:
-                    filter_args = {'object-uri': resource_uri}
-                    resource = cpc.adapters.find(**filter_args)
-                    break
-                except NotFound:
-                    pass  # Try next CPC
-            else:
+            adapters = self.client.consoles.console.list_permitted_adapters(
```
simon-spinner, line 901: This operation was introduced with 2.16.0 and in classic mode it does not return the adapters of older systems. Do we need a fallback path here?

andy-maier: Very good point. I thought that all our `list_permitted_...()` method have implemented the fallback, but that is not the case. Needs to be done here.

andy-maier: Done.

zhmcclient/_console.py
```diff
@@ -990,6 +990,8 @@ def list_permitted_adapters(
         HMC/SE version requirements:
 
         * HMC version >= 2.16.0
+        * The adapters of CPCs with SE version < 2.16.0 are not included in
```
simon-spinner, line 993:
```suggestion
        * The adapters of CPCs with SE version < 2.16.0 and not enabled for DPM are not included in
```

andy-maier: Changed as proposed, with minor wording change:
"The adapters of CPCs with SE version < 2.16.0 that are not enabled for DPM are not included in the result."
Done.

simon-spinner (approved):

PR #1780, opened 2025-03-21
Add support for specifying all CPCs in HMC inventory file

andy-maier:


tests/end2end/test_hmc_inventory.py
```diff
@@ -25,12 +25,46 @@
...
+        print(f"Found test CPC {cpc.name} ({mode_str} mode)")
+
+
+def test_hmcdef_dpm_cpcs(dpm_mode_cpcs):  # noqa: F811
+    # pylint: disable=redefined-outer-name
+    """
+    Display DPM mode CPCs to be tested.
+    """
+    for cpc in dpm_mode_cpcs:
+        mode_str = 'DPM' if cpc.dpm_enabled else 'classic'
```
EdGue42, line 54: code duplication: three times the same loop body?

EdGue42 (approved):

tests/end2end/test_hmc_inventory.py
andy-maier: The body is two lines long and it prints a line.
Not sure it is worth while to have a function for printing the line.

andy-maier: Added a function `_print_cpc()` that is used in the loops.

PR #1812, opened 2025-04-04
1.20: Build test: Install pyzmq from source tarball

andy-maier:


andy-maier:
The reason it did not fail here is because the GitHub Actions environments have a cmake version preinstalled that is <4.0
It only fails with cmake >=4.0.0, and when cmake is not installed at the OS level, the Python cmake package is used which is also at 4.0.0.

The problem has been fixed by pyzmq, and a new release of pyzmq should have the fix. See https://github.com/zeromq/pyzmq/issues/2087#issuecomment-2778447169

PR #1832, opened 2025-04-15
Clarified that Cpc.get_wwpns() is no longer supported

andy-maier:

Other than suggested in the corresponding issue, this PR does not implement the `Cpc.get_wwpns()` method with a loop through attached storage groups and their VSRs, but instead just clarifies that the method was only supported for SE version 2.13.1. It seems to me that the design of the function (being on the CPC, supporting a set of partitions) does not make it worthwhile to support with an inefficient implementation (loop through partitions, loop through their attached storage groups eliminating duplicates, loop through their VSRs).


GitCop:

 * Commit: c8243567122dec6f99dbb67a0538a3bd54aab544

---

anilak11 (approved):

PR #1834, opened 2025-04-15
Support for Console / CPC Hardware Messages (using resource class)

andy-maier:

I tested it with the new end2end test test_hw_message.py on T224, and verified that the console hardware messages exceed the request limit for "Submit Requests" and are properly split into two operations.


andy-maier:
Since people are busy right now, I am merging this PR without further review.

PR #1935, opened 2025-08-24
Enabled adapters for classic mode CPCs

andy-maier:


zhmcclient/_adapter.py
```diff
@@ -263,6 +236,20 @@ class Adapter(BaseResource):
     """
     Representation of an :term:`Adapter`.
 
+    Note that adapter objects do not correspond 1:1 with the physical adapter
+    cards. Adapter objects do correspond 1:1 with PCHIDs, though.
```
simon-spinner, line 240: Nit-pick: Adapter objects may also correspond to VCHIDs for virtual adapters.

simon-spinner (approved):

PR #1960, opened 2025-09-04
Removed the pinning on typer version <0.17.0 with the release of safety 3.6.1

anilak11:
* typer version was pinned to 0.17.0 due to https://github.com/pyupio/safety/issues/778...and this is fixed with 3.6.1 safety version
* so we no longer need to pin that version


andy-maier (changes requested):
You also need to increase the minimum version of safety to 3.6.1 then (in both the dev-requirements.txt file and in the minimum-constraints-develop.txt file).

andy-maier (approved):

PR #1976, opened 2025-09-17
Removed dependency on python-dateutil

andy-maier:

### DIFF
--- a/minimum-constraints-develop.txt
+++ b/minimum-constraints-develop.txt
@@ -202,6 +202,7 @@ pickleshare==0.7.4
 pkginfo==1.4.2
 pyproject-api==1.6.1  # used by tox since its 4.0.0
 pyrsistent==0.20.0  # used by jsonschema>3.0,<4.18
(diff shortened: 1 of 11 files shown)


minimum-constraints-develop.txt
```diff
@@ -202,6 +202,7 @@ pickleshare==0.7.4
 pkginfo==1.4.2
 pyproject-api==1.6.1  # used by tox since its 4.0.0
 pyrsistent==0.20.0  # used by jsonschema>3.0,<4.18
+python-dateutil==2.8.2
```
EdGue42, line 205: if you python-dateutil from the deps, why does it show up here now?

EdGue42 (approved):

minimum-constraints-develop.txt
andy-maier: Because other dev packages need it (e.g. jupyter_client), and we want to reference all direct or indirect dependencies in the minimum-constraints files.

PR #2121, opened 2026-03-12
Improved logging for end2end testcases

andy-maier:
For details, see the commit mesage.

### DIFF
--- a/tests/end2end/conftest.py
+++ b/tests/end2end/conftest.py
@@ -30,6 +30,18 @@ LOG_DATETIME_FORMAT = '%Y-%m-%d %H:%M:%S %Z'
 LOG_DATETIME_TIMEZONE = time.gmtime
 
 
+def has_file_handler(logger, log_file):
+    """
+    Returns whether a Python logger has a FileHandler for the specified log
+    file.
+    """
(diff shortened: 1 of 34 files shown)


tests/end2end/conftest.py
```diff
@@ -54,27 +66,24 @@ def zhmc_logger(request):
...
-
-    if log_file:
+        if not has_file_handler(testfunc_logger, log_file):
+            testfunc_logger.addHandler(log_handler)
         testfunc_logger.setLevel(logging.DEBUG)
     else:
         testfunc_logger.setLevel(logging.NOTSET)
 
-    testfunc_logger.debug("Entered test function")
+    testfunc_logger.debug("Entered test function for %s", request.node.name)
```
EdGue42, line 85: wondering: would it be more consistent to use f-strings everywhere?

EdGue42 (approved):

tests/end2end/conftest.py
andy-maier: Python logging wants % formatting. You get a pylint warning if you use f-strings. The reason is that with % formatting, the construction of the final string can be deferred until after the check whether the log record actually needs to be written (e.g. based on the log level),

PR #2139, opened 2026-04-12
In Makefile on Windows: Used bash and cleaned up Windows-specific commands

andy-maier:

Needs to be applied to all projects.
Should be rolled back to stable version to avoid future merge conflicts.


andy-maier:
Since Chaithra approved PR #2138 which at the time of approval contained this PR, I am considering this PR approved as well.

PR #2198, opened 2026-06-18
Added check for missing and extra dependencies based on installed packages

andy-maier:


.github/workflows/test.yml
```diff
@@ -54,6 +54,11 @@ jobs:
                 \"python-version\": \"3.14\", \
                 \"package_level\": \"latest\" \
               }, \
+              { \
+                \"os\": \"windows-latest\", \
+                \"python-version\": \"3.12\", \
```
simon-spinner, line 59: Why did you pick a python version in the middle of the supported range?

.github/workflows/test.yml
```diff
@@ -273,6 +294,57 @@ jobs:
...
+    - name: Download extra_reqs_install artifacts
+      uses: actions/download-artifact@v8
+      with:
+        pattern: extra_reqs_install_*
+        path: extra_reqs_install/
+        merge-multiple: true
+    - name: Display extra dependencies in minimum-constraints-install.txt
+      shell: bash
+      run: |
+        files=(extra_reqs_install/*.txt); \
```
simon-spinner, line 306: Could we reuse the checks in the make files here in some way?

EdGue42 (commented):
I am wondering, shouldn't the new make file targets be mentioned in the developer documentation?

.github/workflows/test.yml
andy-maier: Because that version on windows has some special package dependencies that made problems in the past:
* pywin32
* pywinpty

.github/workflows/test.yml
andy-maier: You mean for the other projects? That would require writing a github action with the reused code. I started that, but it did not work right away, so for now I suggest to go with repeating the code in all projects, and if and when the action comes to live, we can use that and simplify the makefiles.

andy-maier:
Edwin, we are not documenting each make target in the developer documentation. We are just saying:
> Generally, this project uses Make to do things in the currently active
Python environment. The command ``make help`` (or just ``make``) displays a
list of valid Make targets and a short description of what each target does.

PR #2206, opened 2026-06-28
Added support for Storage Site APIs (1)

### LINKED ISSUE #1675: Support resources types for FICON support in DPM mode
### Storage Site
* new resource; all operations

### Storage Fabric
* new resource; all operations

### Storage Switch
* new resource; all operations

### Storage Subsystem
* new resource; all operations

### Storage Control Unit
* new resource; all operations

### LINKED ISSUE #2248: Support for storage sites
Add support for storage site objects, as child objects of the Console object. The usual methods for creatable objects should be supported.

Add mock support for storage site objects. The storage site objects for the primary site should be predefined (in the mock definition files, not in the code) and should be non-deletable, consistent with the behavior of a real HMC.

Add mock definitions for the predefined storage sites for each CPC.

Add testcases for the nock support.

vontedduchaithra:


tests/end2end/test_storage_site.py
```diff
@@ -0,0 +1,439 @@
+# Copyright 2025 IBM Corp. All Rights Reserved.
```
andy-maier: Please change the copyright year to the current year.

docs/appendix.rst
```diff
@@ -527,6 +527,15 @@ Resources scoped to CPCs in DPM mode
   Storage Port
      Short term for a :term:`Port` of a :term:`Storage Adapter`.
 
+  Storage Site
+     A location that houses a set of storage switches and storage subsystems
+     within the FICON storage configuration of a DPM-enabled :term:`CPC`.
+     A primary site (named "Primary Site") exists by default and cannot be
+     deleted. An alternate site at a remote location can be created and
+     deleted.
```
andy-maier: The existing test does not distinguish sufficiently between the real world storage site and the HMC object representing it.

Also, it seems to suggest that the way one can detect the storage site object for the primary site is by `name="Primary Site"`, when it actually should be detected by `type="primary"`. Also, the `name` property is writeable, so the name is not a reliable way to detect the primary site object.

To address both of that, I suggest to change the two sentences starting at "A primary site (named "Primary Site") exists by default" to:

```
A Storage Site object with `type="primary"` always exists for the primary
site and cannot be deleted. Additional Storage Site objects can be created
and deleted.
```

tests/unit/zhmcclient/test_storage_site.py
```diff
@@ -0,0 +1,294 @@
+# Copyright 2025 IBM Corp. All Rights Reserved.
```
andy-maier: Please change the copyright year to the current year.

zhmcclient/_storage_site.py
```diff
@@ -0,0 +1,318 @@
+# Copyright 2025 IBM Corp. All Rights Reserved.
```
andy-maier: Please change the copyright year to the current year.

zhmcclient/_storage_site.py
```diff
@@ -0,0 +1,318 @@
...
+# limitations under the License.
+
+"""
+A :term:`Storage Site` represents a single storage site in the FICON storage
+configuration associated with a DPM-enabled CPC.
+
+A storage site describes a location that houses a set of storage switches and
+storage subsystems. A primary site with a default name of "Primary Site",
+local to the CPC, exists by default and cannot be deleted. An alternate site,
+typically at a remote location, can be created.
```
andy-maier: The existing test does not distinguish sufficiently between the real world storage site and the HMC object representing it.

Also, it seems to suggest that the way one can detect the storage site object for the primary site is by `name="Primary Site"`, when it actually should be detected by `type="primary"`. Also, the `name` property is writeable, so the name is not a reliable way to detect the primary site object.

To address both of that, I suggest to change the two sentences starting at "A primary site (named "Primary Site") exists by default" to:

```
A Storage Site object with `type="primary"` always exists for the primary
site and cannot be deleted. Additional Storage Site objects can be created
and deleted.
```

zhmcclient/_storage_site.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def __init__(self, console):
+        # This function should not go into the docs.
+        # Parameters:
+        #   console (:class:`~zhmcclient.Console`):
+        #     Console defining the scope for this manager.
+
+        # Resource properties that are supported as filter query parameters.
+        query_props = [
+            'cpc-uris',
```
andy-maier, line 69: The query parameter for the `cpc-uris` property is named `cpc-uri`.
So far, the assumption in the code is that the property has the same name as the query parameter.

That is not the case here, so you need to add special handling of that to the code, and also make sure there is a testcase for that case. The property name here should probably remain `cpc-uris` because that will be the property that is specified in the `filter_args` parameter of `list()`. The special handling is probably in `_list_with_operation()` or in `divide_filter_args()`, but these are very generic methods/functions and you don't want to hard code that case there. I think it requires some more class variable in the StorageSiteManager object to provide the necessary data to these functions. You need to work out how you want to handle that, and we can discuss it in the weekly meeting.

zhmcclient/_storage_site.py
```diff
@@ -0,0 +1,318 @@
...
+            'name',
```
andy-maier, line 70: There is an additional supported query property: `type`.

andy-maier (changes requested):

tests/end2end/test_storage_site.py
vontedduchaithra: Resolved

docs/appendix.rst
vontedduchaithra: Resolved

tests/unit/zhmcclient/test_storage_site.py
vontedduchaithra: Resolved

zhmcclient/_storage_site.py
vontedduchaithra: Resolved

zhmcclient/_storage_site.py
vontedduchaithra: Resolved

zhmcclient/_storage_site.py
vontedduchaithra: Resolved

andy-maier (dismissed):

andy-maier:
I approved the PR, but the merge conflict needs to be resolved and the commits shoul dbe quashed into a single one that has a reasonable description.

zhmcclient/_storage_site.py
andy-maier: I'm fine with the resolution.

andy-maier (approved):

PR #2207, opened 2026-07-01
Bump from jupyter_server >=2.10.0 to jupyter_server >=2.20.0 

vontedduchaithra:
Fixes path traversal vulnerability in jupyter_server <=2.17.0.

### DIFF
--- a/changes/37.fix.rst
+++ b/changes/37.fix.rst
@@ -1,0 +1,6 @@
+Bumped ``jupyter_server`` to >=2.20.0 on Python>=3.10 to fix
+CVE-2024-35178 (GHSA-hrw6-wg82-cm62), a path traversal vulnerability
+caused by an incorrect ``startswith()`` root-directory check that
+allowed authenticated users to read, write, and delete files in sibling
+directories sharing a name prefix with the server's ``root_dir``.
+Also documented this Dependabot alert pattern in the development guide.
--- a/changes/noissue.37.fix.rst
+++ b/changes/noissue.37.fix.rst
@@ -3,4 +3,3 @@ CVE-2024-35178 (GHSA-hrw6-wg82-cm62), a path traversal vulnerability
 caused by an incorrect ``startswith()`` root-directory check that
 allowed authenticated users to read, write, and delete files in sibling
 directories sharing a name prefix with the server's ``root_dir``.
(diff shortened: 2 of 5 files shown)


changes/noissue.37.fix.rst
```diff
@@ -0,0 +1,6 @@
+Bumped ``jupyter_server`` to >=2.20.0 on Python>=3.10 to fix
```
andy-maier, line 1: The number in the filename (37) must be the issue number you are addressing, because towncrier creates a link to this issue in the generated change log.

If there is no corresponding issue (like in this case), use `noissue.NN.fix.rst` as the filename, where `NN` is some number that is not yet used in this release (like 37). The `noissue` text at the start of the filename causes towncrier not to create an issue link.

changes/37.fix.rst
```diff
@@ -0,0 +1,6 @@
+Bumped ``jupyter_server`` to >=2.20.0 on Python>=3.10 to fix
+CVE-2024-35178 (GHSA-hrw6-wg82-cm62), a path traversal vulnerability
+caused by an incorrect ``startswith()`` root-directory check that
+allowed authenticated users to read, write, and delete files in sibling
+directories sharing a name prefix with the server's ``root_dir``.
+Also documented this Dependabot alert pattern in the development guide.
```
andy-maier: I did not see a change in the development guide for that in this PR.
The handling of Dependabot alerts is already documented in section "Handling Dependabot alerts and PRs". If you want to change that, I suggest to do that in a separate PR.

If there is no change in the development guide in this PR, the respective sentence should be removed from the change log.

andy-maier (changes requested):

changes/noissue.37.fix.rst
vontedduchaithra: Resolved

changes/37.fix.rst
vontedduchaithra: Resolved

andy-maier (approved):

andy-maier:
Note that PR #2210 already had increased the version of jupiter_server to 2.20, because stating 2.10 was actually a mistake in an earlier PR. So this PR got reduced to just providing the change log entry, upon its merge.

PR #2224, opened 2026-07-10
Support for additional_properties in StorageVolumeManager.list()

andy-maier:

### DIFF
--- a/tests/end2end/test_storage_volume.py
+++ b/tests/end2end/test_storage_volume.py
@@ -87,7 +87,7 @@ def test_stovol_find_list(zhmc_logger, dpm_mode_cpcs):
             runtest_find_list(
                 session, stogrp.storage_volumes, stovol.name, 'name', 'size',
                 STOVOL_VOLATILE_PROPS, STOVOL_MINIMAL_PROPS, STOVOL_LIST_PROPS,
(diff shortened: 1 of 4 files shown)


tests/end2end/test_storage_volume.py
```diff
@@ -87,7 +87,7 @@ def test_stovol_find_list(zhmc_logger, dpm_mode_cpcs):
             runtest_find_list(
                 session, stogrp.storage_volumes, stovol.name, 'name', 'size',
                 STOVOL_VOLATILE_PROPS, STOVOL_MINIMAL_PROPS, STOVOL_LIST_PROPS,
-                unique_name=unique_name)
+                add_props=['uuid'], unique_name=unique_name)
```
EdGue42, line 90: I don't understand the syntax here ... is "add_props" a named parameter on runtest_find_list?

Because on the actual method in the handler, the parameter is named "additional_properties"?

andy-maier: add_props is indeed a parameter of runtest_find_list(): https://github.com/zhmcclient/python-zhmcclient/blob/master/tests/end2end/utils.py#L158

EdGue42 (approved):

PR #2227, opened 2026-07-14
Added mock definitions for SSO Server Definitions

### LINKED ISSUE #2124: Add mock definitions for SSO Server Definitions
We have recently added support for SSO server definitions including mock support.

This ticket is for adding at least one SSO server definition object to the mock yaml files. Since SSO Server Definition support was added in z17, this requires first to copy the mock yaml file for z16 to z17, and then add it only in the file for z17.

vontedduchaithra:


tests/end2end/test_sso_server_definition.py
```diff
@@ -189,7 +189,12 @@ def test_ssosrvdef_crud(zhmc_logger, hmc_session):
             ssosrvdef.properties[pn] == exp_value
         ), f"Unexpected value for property {pn!r}"
     ssosrvdef.pull_full_properties()
+    # After GET, server-controlled (auto) properties override the input values.
```
andy-maier: This has been fixed in PR #2226  - I suggest that you rebase this PR on that PR and check whether it works.

andy-maier (changes requested):

tests/end2end/test_sso_server_definition.py
vontedduchaithra: Done

andy-maier (approved):

PR #2240, opened 2026-07-29
Changed StorageGroup.list_candidate_adapter_ports() to return None for FC-type

### LINKED ISSUE #2239: StorageGroup.list_candidate_adapter_ports() raises KeyError for FC-type storage groups
FC-type storage groups do not have a property "candidate-adapter-port-uris". The `StorageGroup.list_candidate_adapter_ports()` method accesses that property unconditionally and raises `KeyError` for FC-type storage groups.

While the method is documented to apply only to FCP-type storage groups, it leaves it open what happens when called for FC-type storage groups. The `KeyError` that is raised is not documented.

Proposal: The method should return `None` in case of an FC-type storage group, and document that.

andy-maier:

### DIFF
--- a/zhmcclient/_storage_group.py
+++ b/zhmcclient/_storage_group.py
@@ -674,7 +674,9 @@ class StorageGroup(BaseResource):
         Return the current candidate storage adapter port list of this FCP
         storage group.
 
-        This operation only applies to storage groups of type "fcp".
+        This operation only returns candicate adapter ports for storage groups
+        of type "fcp".
+        When called for storage groups of type "fc", it returns `None`.
 
         The result reflects the actual list of ports used by the CPC, including
         any changes that have been made during discovery. The source for this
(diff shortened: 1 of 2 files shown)


vontedduchaithra (approved):

zhmcclient/_storage_group.py
```diff
@@ -674,7 +674,9 @@ def list_candidate_adapter_ports(self, full_properties=False):
         Return the current candidate storage adapter port list of this FCP
         storage group.
 
-        This operation only applies to storage groups of type "fcp".
+        This operation only returns candicate adapter ports for storage groups
+        of type "fcp".
+        When called for storage groups of type "fc", it returns `None`.
```
simon-spinner, line 679: What about storage groups of type nvme? Should be included here or generally state for other types...

PR #2246, opened 2026-07-30
Removed click from requirements-develop.txt because it is not a direct dependency

andy-maier:
No review needed.


andy-maier:
Note: Does not need to be backported to 1.26 because click is not a dependency in requirements-develop.txt in that version.

PR #2253, opened 2026-08-12
Added support for Storage Switch APIs (2)

### LINKED ISSUE #2250: Support for storage switches
Add support for storage switch objects, as child objects of the Console object. The usual methods for creatable objects should be supported.

The standard list() method should be implemented by invoking one of the two HMC list operations ("list switches of a fabric", "list switches of a site"), dependent on which filter arguments are provided: If storage-fabric-uri is provided, use "list switches of a fabric". Else if storage-site-uri is provided, use "list switches of a site". Else, list storage sites (with a CPC filter, if provided to the list() method) and for each site, list the switches of the site using "list switches of a site".

In addition, the two move operations should be supported with special methods on the storage switch object: move_to_site(), move_to_fabric().

Add mock definitions for some storage switches for some CPCs.

vontedduchaithra:


andy-maier (changes requested):
Please decouple your PRs. See my comment on #2254 for details.

andy-maier (approved):

PR #2254, opened 2026-08-12
Added support for Storage Subsystem APIs (3)

### LINKED ISSUE #2251: Support for storage subsystems
Add support for storage subsystem objects, as child objects of the Console object. The usual methods for creatable objects should be supported.

The standard list() method should be implemented by invoking the HMC list operation "list subsystems of a site", dependent on which filter arguments are provided: If the artificial filter argument storage-site-uri is provided, use "list subsystems of a site" using that URI. Else, list storage sites (with a CPC filter, if provided to the list() method) and for each site, list the subsystems of the site using "list subsystems of a site".

Add support for the move operation via a special method move_to_site().

Add support for the add/remove connection endpoint operations via special methods that are separate for each endpoint object type: add/remove_connection_adapter(), add/remove_connection_switch_port().

Add mock definitions for some storage subsystems for some CPCs.

vontedduchaithra:


andy-maier (changes requested):
The PR contains both storage site and storage subsystem, and given the 8 commits it is hard to find out what belongs to storage subsystem.

If you need storage site as a basis for storage subsystem, then please squash each of them into a single commit. I.e. you first suqash the 4 commits in PR #2206 into a single one, and then rebase this PR on that PR and squash the remaining commits of this PR into a single one, so that the resulting PR has two commits.

If you do not need storage site as a basis, then simply remove its commits from this PR and squash the remaining commits into a single one.

andy-maier (changes requested):
Please rebase the PR on the latest master and resolve the merge conflicts.

andy-maier:
Please rebase this PR to the latest master, squash the commits into a single one, and resolve the merge conflicts.

andy-maier (approved):

PR #2255, opened 2026-08-13
Added support for Storage Control Unit APIs (4)

### LINKED ISSUE #2252: Support for storage control units, including storage paths
Add support for storage control unit objects, as child objects of the storage subsystem (!) object. The usual methods for creatable objects should be supported.

The standard list() method can be implemented straight forward by using the HMC list operation "list control units of a subsystem".

Add support for the add/remove volume range operations via special methods: add/remove_volume_range().

Add support for the storage path element object as a standard child object of the control unit object via attribute "storage_paths".

The storage path objects should support the standard methods for creatable element objects:
* list() based on the storage-path-uris property
* create()/delete()/update_properties() using the corresponding API operations.
* the usual other methods/attributes.

Add mock definitions for some storage control units for some CPCs.

vontedduchaithra:


andy-maier (changes requested):
Please decouple your PRs. See my comment on #2254 for details.

andy-maier (changes requested):
Please rebase on latest master and resolve the merge conflicts.

